PR #109, opened 2023-04-06
feat(bignumber): this will solve the type issues on extension repo

vaske:


teebszet:
@yknl can we merge this? I just rebased onto origin/develop

PR #110, opened 2023-04-12
Import account, create & sign TX with Ledger

normanwilde:
WIP

- Added functions to pull BTC and Ordinals addresses from the Ledger device
- Added functions to create a PSBT based on arguments received, send the PSBT to the Ledger device and retrieve it signed (ready to be broadcast with the existing broadcasting function)


m-aboelenein:
closed since it's no longer relevant

PR #159, opened 2023-06-13
filter unspent outputs using ordinals utxos

m-aboelenein:
`signOrdinalSendTransaction` function is updated to take the a list of the address ordinals utxo to filter them out from the unspent outputs used to construct the transfer tx

### DIFF
--- a/transactions/btc.ts
+++ b/transactions/btc.ts
@@ -676,10 +676,7 @@ export async function signOrdinalSendTransaction(
   // This can be true if ordinal utxo is from the payment address
 
 
-  const filteredUnspentOutputs = [
-    ...filterUtxos(unspentOutputs, addressOrdinalsUtxos),
-    ...filterUtxos(addressOrdinalsUtxos, unspentOutputs),
-  ];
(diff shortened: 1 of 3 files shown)


transactions/btc.ts
```diff
@@ -675,8 +677,8 @@ export async function signOrdinalSendTransaction(
 
 
   const filteredUnspentOutputs = [
-    ...filterUtxos(unspentOutputs, addressOrdinalsUtxos!),
-    ...filterUtxos(addressOrdinalsUtxos!, unspentOutputs),
+    ...filterUtxos(unspentOutputs, addressOrdinalsUtxos),
+    ...filterUtxos(addressOrdinalsUtxos, unspentOutputs),
```
yknl, line 681: Doesn't this just give you `unspentOutputs` + `addressOrdinalsUtxos`? Which means the ordinals would still be in the UTXO set

m-aboelenein: it does the opposite actually so it filters the first arr wit the content of the second one and then filters the second arr with the content of the first one the only way this could break is if the user have more ordinals utxos than utxos all-together which is not possible but basically the second filter is redundant

yknl (approved):

PR #162, opened 2023-06-16
Fix incorrect BTC fee

abdulhaseeb4239:
The fee would be different every time we fetch unspent outputs and calculate the fee from TX size.

I have added the sorting of unspent outputs before selecting for a transaction, on the basis of block time. This fixes the inconsistent fee issue. 

The unit tests are updated with correct mocking data. 

```
// Sort UTXOs based on block time in ascending order
  unspentOutputs.sort((a, b) => {
    if (a.status.block_time && b.status.block_time) {
      return a.status.block_time - b.status.block_time;
    }
    // Handle cases where block time is missing
    return 0;
  });
```

### DIFF
--- a/tests/btc/transactions.test.ts
+++ b/tests/btc/transactions.test.ts
@@ -562,19 +562,6 @@ describe('bitcoin transactions', () => {
           confirmed: true,
         },
         txid: '8b330459af5329c06f8950fda313bbf2e51afc868e3b31c0e1a7acbca2fdffe6',
-        value: 5700,
-        vout: 0,
-      },
-      {
-        address: '3Codr66EYyhkhWy1o2RLmrER7TaaHmtrZe',
-        blockHeight: 793556,
-        status: {
-          block_hash: '00000000000000000000a46de80f72757343c538d13be3a992aa733fe33bc4bb',
-          block_height: 793556,
-          block_time: 1686310361,
-          confirmed: true,
-        },
-        txid: '8b330459af5329c06f8950fda313bbf2e51afc868e3b31c0e1a7acbca2fdffe6',
         value: 3911,
         vout: 1,
       },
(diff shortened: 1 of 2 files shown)


tests/btc/transactions.test.ts
```diff
@@ -724,7 +706,8 @@ describe('bitcoin transactions', () => {
 
     const filteredUnspentOutputs = utxos.filter((unspentOutput) => {
```
dhriaznov, line 707: Should we reuse the `filterUtxos` function from `transactions/btc.ts` file here?

transactions/btc.ts
```diff
@@ -65,6 +65,15 @@ export function selectUnspentOutputs(
     sumValue += pinnedOutput.value;
   }
 
+  // Sort UTXOs based on block time in ascending order
+  unspentOutputs.sort((a, b) => {
```
m-aboelenein, line 69: we can sort and calculate the sum in one `Arr.reduce` statement instead of looping twice here

yknl:
Please also fix the tests @abdulhaseeb4239

transactions/btc.ts
victorkirov: Yeah, I don't think that's possible 👀  We'd need to select from the sorted list.

transactions/btc.ts
```diff
@@ -65,6 +65,15 @@ export function selectUnspentOutputs(
...
+    if (a.status.block_time && b.status.block_time) {
+      return a.status.block_time - b.status.block_time;
+    }
+    // Handle cases where block time is missing
+    return 0;
```
victorkirov, line 74: I don't think we should return a 0 here. It could result in things going out of order again and causing similar issues and it would also result in things without a block time randomly being placed in the ordered UTXOs that have a block time.

Instead, we should do something like:
```javascript
if (a.status.block_time) {
  return 1;
} 
if (a.status.block_time) {
  return -1;
} 

return a.amount - b.amount; // If neither have block time, then place them at the end of the list ordered by amount so we use UTXOs with higher amount first
```

victorkirov (changes requested):

tests/btc/transactions.test.ts
```diff
@@ -161,58 +162,45 @@ describe('bitcoin transactions', () => {
     const unspent3Value = 250000;
     const totalUnspentValue = unspent1Value + unspent2Value + unspent3Value;
 
-    const utxos: Array<UTXO> = [
+    const utxos: Array<UTXO & { address: string; blockHeight?: number }> = [
```
victorkirov, line 165: This test seems like it was changed quite a bit and there are variables in it that are no longer used; it also doesn't seem to be testing the function as thoroughly as it used to 🤔 Should it not still have worked with a slight tweak to take into account the different UTXO order? I think we should put it back to what it was and try figure out if it legitimately broke and we need to fix the logic or try tweak the test to fix it for the new logic.

tests/btc/transactions.test.ts
```diff
@@ -296,61 +284,52 @@ describe('bitcoin transactions', () => {
...
+        value: 247106,
+        vout: 1,
       },
     ];
 
-    const recipient1Amount = 200000;
+    const recipient1Amount = 6000;
     const recipient2Amount = 100000;
-    const satsToSend = recipient1Amount + recipient2Amount;
+    const satsToSend = recipient1Amount;
```
victorkirov, line 318: The same with this test. We should try and fix it rather than rewrite it. e.g. recipient2Amount amount is still defined but is never used. We should keep it and try fix the test.

victorkirov (approved):
LGTM 🙂

PR #168, opened 2023-06-28
Filter utxos in send ordinal get fee function

abdulhaseeb4239:
- added the filterutxos in the fee calculation for ordinal send to make sure ordinal UTXO is removed from UTXO set used for fees. 
- added a try-catch block in `signOrdinalSendTransaction` to catch and reject the promise. 
- updated test for `getBtcFeesForOrdinalSend` function

Change on apps: 
 - we need to pass in an additional parameter addressOrdinalsUtxos added in `getBtcFeesForOrdinalSend` same as `signOrdinalSendTransaction`

### DIFF
--- a/tests/btc/transactions.test.ts
+++ b/tests/btc/transactions.test.ts
@@ -248,6 +248,17 @@ describe('bitcoin transactions', () => {
         },
         value: ordinalValue,
       },
+      {
+        status: {
+          block_hash: '00000000000000000003e6c56ae100b34fcc2967bc1deb53de1a4b9c29ba448f',
+          block_height: 797404,
+          block_time: 1688626274,
+          confirmed: true,
+        },
+        txid: 'd0dfe638a5be4f220f6435616edb5909a2f93540a7d6975ed0bdf305fb8bf51c',
+        value: 1347,
+        vout: 0,
+      },
     ];
 
     const utxos: Array<UTXO> = [
@@ -262,6 +273,7 @@ describe('bitcoin transactions', () => {
         },
         value: unspent1Value,
       },
+      ...ordinalOutputs,
     ];
 
     const fetchFeeRateSpy = vi.spyOn(XverseAPIFunctions, 'fetchBtcFeeRate');
@@ -279,8 +291,6 @@ describe('bitcoin transactions', () => {
 
     const ordinalUtxos = [
       {
-        address: '3BMxVoc3NVt8BHakAh28WZrpQqQKCxV28U',
-        blockHeight: 797404,
         status: {
           block_hash: '00000000000000000003e6c56ae100b34fcc2967bc1deb53de1a4b9c29ba448f',
           block_height: 797404,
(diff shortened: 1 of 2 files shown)


transactions/btc.ts
```diff
@@ -658,88 +659,88 @@ export async function signOrdinalSendTransaction(
   fee?: BigNumber,
 ): Promise<SignedBtcTx> {
   // Get sender address unspent outputs
-  const btcClient = new BitcoinEsploraApiProvider({
-    network,
-  });
-  const unspentOutputs = await btcClient.getUnspentUtxos(btcAddress);
+  try {
```
m-aboelenein, line 662: we should aim to throw descriptive errors throughout this function to handle them on the app-side like when we get the `feeRate` or when we select the outputs for the tx

victorkirov: Could we rather get rid of the try catch block? The function should be throwing an error and not a string, as it is currently doing. If we make that change, then the try/catch block is redundant here.

transactions/btc.ts
```diff
@@ -659,125 +659,122 @@ export async function signOrdinalSendTransaction(
...
+    fee: fee ?? calculatedFee,
+    feePerVByte,
+    total: satsToSend,
+  };
 
-    return await Promise.resolve(signedBtcTx);
-  } catch (error) {
-    return Promise.reject(error.toString());
-  }
+  return await Promise.resolve(signedBtcTx);
```
victorkirov, line 777: Sorry, last one 😅 Could you change this to `return signedBtcTx`? We're doing so many unnecessary awaits and Promise wrapping in this file 😋

abdulhaseeb4239: we would need to do these improvements in all of the btc.ts file. I guess that's for a separate PR. But its a start nonetheless 😄 I'll make this change

tests/btc/transactions.test.ts
```diff
@@ -277,9 +277,29 @@ describe('bitcoin transactions', () => {
     const ordinalAddress = 'bc1prtztqsgks2l6yuuhgsp36lw5n6dzpkj287lesqnfgktzqajendzq3p9urw';
     const btcAddress = '1H8voHF7NNoyz76h9s6dZSeoypJQamX4xT';
 
-    const { fee } = await getBtcFeesForOrdinalSend(recipientAddress, ordinalOutputs[0], btcAddress, network, [
+    const ordinalUtxos = [
```
victorkirov, line 280: This won't really test the filter though because this UTXO is not in the mock list of all UTXOs for the address.

Could you try make these changes so that we actually test the filtering 🙏 :
- add this new UTXO to ordinalOutputs ion line 239 above
- on line 253, the utxo mock data, change it to:
```
const utxos: Array<UTXO> = [
      {
        txid: '1f2bbb92a74d379db2502e8ae7a57917041db5dc531ef54e64ca532aa9f59d8c',
        vout: 2,
        status: {
          confirmed: true,
          block_height: 123123,
          block_time: 1677048365,
          block_hash: '000000000000000000072266ee093771d806cc9cb384461841f9edd40b52b67f',
        },
        value: unspent1Value,
      },
     ...ordinalOutputs,
    ];
```

The resulting fee should remain the same because those 2 UTXOs should get filtered out.

transactions/btc.ts
victorkirov: Thanks :)

victorkirov (approved):

PR #169, opened 2023-07-03
 seed vault

m-aboelenein:
# 🔘 PR Type

This PR introduces the usage of the seed vault a class that manages the storage and retrieval of the seed phrase in the web extension and the mobile app

- [ ] Bugfix
- [x] Enhancement
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes


tests/wallet/encryptDecryptMnemonic.test.ts
```diff
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { decryptMnemonicWithCallback, encryptMnemonicWithCallback } from '../../wallet';
+import { decryptMnemonicWithCallback, encryptMnemonicWithCallback } from '../../seedVault/encryptionUtils';
```
victorkirov, line 3: This file should probably move to the seedvault dir since it's only testing that and nothing specific to the wallet.

seedVault/encryptionUtils.ts
```diff
@@ -0,0 +1,106 @@
+import { mnemonicToEntropy, wordlists } from 'bip39';
+
+interface EncryptMnemonicArgs {
+  password: string;
+  seed: string;
+  mnemonicEncryptionHandler: (seed: string, key: string) => Promise<Buffer>;
+}
+
+export async function encryptMnemonicWithCallback(cb: EncryptMnemonicArgs) {
```
victorkirov, line 9: this naming is a bit confusing because we don't actually do a callback 🤔 Maybe it should be something like `encryptMnemonicWithHandler`

seedVault/encryptionUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  const ctStr = cipherSplitted[1];
+  const ctUint8 = Buffer.from(ctStr, 'base64');
+  try {
+    const plainBuffer = await crypto.subtle.decrypt(alg, key, ctUint8);
+    const plaintext = new TextDecoder().decode(plainBuffer);
+    return plaintext;
+  } catch (e) {
+    throw new Error('Decrypt failed');
+  }
+}
```
victorkirov, line 106: Should we maybe export these in the main index file? Otherwise using them requires the user to import form `@secretkeylabs/xverse-core/seedVault/encryptionUtils`

seedVault/index.ts
```diff
@@ -0,0 +1,81 @@
+export interface SecureStorageAdapter {
+  get(key: string): Promise<string>;
+  set(key: string, value: string): Promise<void>;
+}
```
victorkirov, line 4: Should this adapter not have the same structure as the common one where the get could return a null or non-promise value?

seedVault/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    const passphrase = await this._secureStorageAdapter.get(SeedVaultStorageKeys.PASSWORD_HASH);
+    if (!passphrase) throw new Error('passwordHash not set');
+    const encryptedSeed = await this._cryptoUtilsAdapter.encrypt(seed, passphrase);
+    if (!encryptedSeed) throw new Error('Seed not set');
+    this._commonStorageAdapter.set(SeedVaultStorageKeys.ENCRYPTED_KEY, encryptedSeed);
+  };
+
+  getSeed = async (password?: string) => {
+    let passwordHash = await this._secureStorageAdapter.get(SeedVaultStorageKeys.PASSWORD_HASH);
+    if (!passwordHash && password) {
```
victorkirov, line 62: Do we maybe need to throw an error or something if both the passwordHash and password values are empty?

victorkirov: We should also probably have tests for this class with these scenarios 😅

victorkirov (changes requested):
Left a few comments. One other thing is that I'm not sure how the utils fit into the SeedVault class. Do they somehow fit into the encrypt and decrypt functions in the CryptoUtilsAdapter? If so, maybe they should have a similar type to make the reference more concrete.

seedVault/encryptionUtils.ts
yknl: Yes, that is a bit confusing. Was this originally meant to have a callback?

seedVault/encryptionUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  const ivStr = Buffer.from(iv).toString('base64');
+  const alg = { name: 'AES-GCM', iv };
+  const key = await crypto.subtle.importKey('raw', toUint8(Buffer.from(passwordHash)), alg, false, ['encrypt']);
+  const ptUint8 = new TextEncoder().encode(plaintext);
+  const ctBuffer = await crypto.subtle.encrypt(alg, key, ptUint8);
+  const ctStr = Buffer.from(ctBuffer).toString('base64');
+  return `${ivStr}.${ctStr}`;
+}
+
+export const encryptSeed = async (seed: string, passwordHash: string) => {
```
victorkirov, line 53: For consistency
```suggestion
export const encryptSeed = async (seed: string, passwordHash: string): Promise<string> => {
```

seedVault/encryptionUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @param   {String} plaintext - Plaintext to be encrypted.
+ * @param   {String} passwordHash - Password to use to encrypt plaintext.
+ * @returns {String} Encrypted cipherText.
+ *
+ * @example
+ *   const cipherText = await aesGcmEncrypt('my secret text', 'pw');
+ *   aesGcmEncrypt('my secret text', 'pw').then(function(cipherText) { console.log(cipherText); });
+ */
+export async function aesGcmEncrypt(plaintext: string, passwordHash: string): Promise<string> {
```
victorkirov, line 42: aesGcmEncrypt/Decrypt and encryptSeed should probably have some unit tests

seedVault/index.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export interface CryptoUtilsAdapter {
+  encrypt(data: string, password: string): Promise<string>;
+  decrypt(data: string, password: string): Promise<string>;
+  hash(data: string, salt: string): Promise<string>;
+  generateRandomBytes(length: number): string;
+}
+
+export interface SeedVaultConfig {
```
victorkirov, line 13: These are probably better off as types instead of interfaces. Also, in `CryptoUtilsAdapter`, could we make explicit types for the encrypt and decrypt functions and assign those types to the related functions in the encryptionUtils file? That way there is a clear link between them.

seedVault/encryptionUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  password: string;
+  encryptedSeed: string;
+  mnemonicDecryptionHandler: (seed: Buffer | string, key: string) => Promise<string>;
+}
+
+export async function decryptMnemonicWithHandler(cb: DecryptMnemonicArgs) {
+  const { mnemonicDecryptionHandler, password, encryptedSeed } = cb;
+  const seedPhrase = await mnemonicDecryptionHandler(encryptedSeed, password);
+  return seedPhrase;
+}
```
victorkirov, line 25: I'm not sure how these relate to the seed vault. if they are related, could we make their relationship to the seed vault clear with some sort of typing? If they aren't directly related, maybe they should go back to the wallet section if it's a more appropriate location.

encryption/mnemonic.ts
```diff
@@ -0,0 +1,37 @@
+import { entropyToMnemonic, mnemonicToEntropy, wordlists } from "bip39";
+import { aesGcmDecrypt, aesGcmEncrypt } from "./aes";
+import { decryptMnemonic, encryptMnemonic } from "@stacks/encryption";
+
+
+/**
+ * @param {string} seed
+ * @returns {string}
+ */
+const seedToEntropy = (seed: string): string => {
```
victorkirov, line 10: nit: maybe mnemonicToEntropy would be a better name. Or seedPhrase

encryption/mnemonic.ts
```diff
@@ -0,0 +1,37 @@
...
+try {
+  return mnemonicToEntropy(seed, wordlists.EN);
+} catch (error) {
+  console.error('Invalid mnemonic phrase provided');
+  console.error(error);
+  throw new Error('Not a valid bip39 mnemonic');
+}
+}
+
+export const encryptSeed = async (seed: string, passwordHash: string): Promise<string> => {
```
victorkirov, line 20: Same here. Either menmonic or seedPhrase

encryption/mnemonic.ts
```diff
@@ -0,0 +1,37 @@
...
+export const decryptSeed = async (encryptedSeed: string, passwordHash: string): Promise<string> => {
+  const mnemonicEntropy = await aesGcmDecrypt(encryptedSeed, passwordHash);
+  const seed = entropyToMnemonic(mnemonicEntropy, wordlists.EN);
+  return seed;
+};
+
+
+export const encryptSeedCBC = encryptMnemonic;
+
+export const decryptSeedCBC = decryptMnemonic;
```
victorkirov, line 37: Do we need to export these? Also, what does CBC stand for? Could we rather use the full version if it's an abbreviation?

index.ts
```diff
@@ -8,4 +8,5 @@ export * from './account';
 export * from './gaia';
 export * from './ledger';
 export * from './coins';
-export * from './seedVault';
+export { default as SeedVault } from './seedVault';
```
victorkirov, line 11: This won't export the interfaces/types and the enum. I think you also need `export * from './seedvault'`

utils/arrayBuffers.ts
```diff
@@ -0,0 +1,33 @@
+export const toUint8 = (buf: Buffer): Uint8Array => {
+  const uin = new Uint8Array(buf.length);
+  return uin.map((a, index, arr) => (arr[index] = buf[index]));
```
victorkirov, line 3: I'm not sure we really need this method since you can convert it in one line. If we keep it, could you rather change it to this?

```suggestion
return new Uint8Array(buf);
```

utils/arrayBuffers.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export function concatBuffers(buf1: ArrayBuffer, buf2: ArrayBuffer) {
+  const result = new Uint8Array(buf1.byteLength + buf2.byteLength);
+  result.set(new Uint8Array(buf1), 0);
+  result.set(new Uint8Array(buf2), buf1.byteLength);
+  return result.buffer;
```
victorkirov, line 10:
```suggestion
return new Uint8Array([...buf1, ...buf2]);
```

utils/arrayBuffers.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Encodes a byte array as a string of hex.
+ * @param {Uint8Array} buffer
+ * @returns {String}
+ */
+export function buf2hex(buffer: ArrayBuffer): string {
+  return Array.prototype.slice
+    .call(new Uint8Array(buffer))
+    .map(x => [x >> 4, x & 15])
+    .map(ab => ab.map(x => x.toString(16)).join(""))
+    .join("");
```
victorkirov, line 32:
```suggestion
return Buffer.from(buffer).toString('hex');
```

encryption/aes.ts
```diff
@@ -0,0 +1,44 @@
...
+export async function aesGcmDecrypt(cipherText: string, passwordHash: string): Promise<string> {
+  const cipher = Buffer.from(cipherText, 'hex');
+  const data = cipher.subarray(IV_SIZE);
+  const iv = cipher.subarray(0, IV_SIZE);
+  const alg = { name: 'AES-GCM', iv};
+  const key = await crypto.subtle.importKey('raw', str2buf(passwordHash), alg, false, ['encrypt', 'decrypt']);
+  const plainBuffer = await crypto.subtle.decrypt(alg, key, data);
+  const plaintext = new TextDecoder().decode(plainBuffer);
+  return plaintext;
+}
```
victorkirov, line 44: Would be good to have some aesGcmDecrypt and aesGcmEncrypt tests

utils/arrayBuffers.ts
```diff
@@ -0,0 +1,33 @@
...
+}
```
victorkirov, line 33: Maybe best to add unit tests for this file and ensure that the suggested code changes match the old functionality. I tested them all manually but also best to have unit tests for this.

encryption/mnemonic.ts
teebszet: mnemonicToEntropy is declared, so maybe seedPhraseToEntropy

seedVault/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+class SeedVault {
+  private readonly _secureStorageAdapter: StorageAdapter;
+
+  private readonly _cryptoUtilsAdapter: CryptoUtilsAdapter;
+
+  private readonly _commonStorageAdapter: StorageAdapter;
+
+  constructor(config: SeedVaultConfig) {
+    this._secureStorageAdapter = config.storageAdapter;
```
teebszet, line 33: could we rename this secureStorageAdapter to be clear?

seedVault/index.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+class SeedVault {
+  private readonly _secureStorageAdapter: StorageAdapter;
+
+  private readonly _cryptoUtilsAdapter: CryptoUtilsAdapter;
+
+  private readonly _commonStorageAdapter: StorageAdapter;
+
+  VERSION = 1;
```
victorkirov, line 32: I think we'll need to store this in common storage so we would know what version the stored data is in when doing an upgrade.

index.ts
victorkirov: @m-aboelenein just re-checking if you considered this

victorkirov (approved):
LGTM!

seedVault/index.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  unlockVault = async (password: string): Promise<string> => {
+    try {
+      const encryptedSeed = await this._commonStorageAdapter.get(SeedVaultStorageKeys.ENCRYPTED_KEY);
+      const salt = await this._commonStorageAdapter.get(SeedVaultStorageKeys.PASSWORD_SALT);
+      if (salt && encryptedSeed) {
+        const passwordHash = await this._cryptoUtilsAdapter.hash(password, salt);
+        const seedPhrase = await this._cryptoUtilsAdapter.decrypt(encryptedSeed, passwordHash as string);
+        await this._secureStorageAdapter.set(SeedVaultStorageKeys.PASSWORD_HASH, passwordHash);
+        return seedPhrase;
```
victorkirov, line 90: This should probably not return the seed phrase. If someone wants the seed phrase as soon as they unlock, they should call:
```
await unlockVault()
seedPhrase = getSeed()
```

seedVault/index.ts
```diff
@@ -0,0 +1,111 @@
...
+      } else {
+        throw new Error('empty vault');
+      }
+    } catch (err) {
+      if (err.message === 'empty vault') {
+        throw err;
+      }
+      throw new Error('Wrong password');
+    }
+  };
```
victorkirov, line 100: This feels a bit cleaner

```suggestion
  unlockVault = async (password: string): Promise<void> => {
    const encryptedSeed = await this._commonStorageAdapter.get(SeedVaultStorageKeys.ENCRYPTED_KEY);
    const salt = await this._commonStorageAdapter.get(SeedVaultStorageKeys.PASSWORD_SALT);
    if (salt && encryptedSeed) {
      try {
        const passwordHash = await this._cryptoUtilsAdapter.hash(password, salt);
        const seedPhrase = await this._cryptoUtilsAdapter.decrypt(encryptedSeed, passwordHash as string);
        await this._secureStorageAdapter.set(SeedVaultStorageKeys.PASSWORD_HASH, passwordHash);
      } catch (err) {
        throw new Error('Wrong password');
      }
    } else {
      throw new Error('empty vault');
    }
  };
```

victorkirov (changes requested):
Sorry, a few changes arose from the extension review 😅

victorkirov (approved):
LGTM 👍

teebszet (approved):

seedVault/index.ts
```diff
@@ -49,9 +49,13 @@ class SeedVault {
     this._secureStorageAdapter.set(SeedVaultStorageKeys.PASSWORD_HASH, passwordHash);
   };
 
-  storeSeed = async (seed: string) => {
+  storeSeed = async (seed: string, overwriteExistingSeed?: boolean) => {
     const password = await this._secureStorageAdapter.get(SeedVaultStorageKeys.PASSWORD_HASH);
     if (!password) throw new Error('passwordHash not set');
+    const prevStoredEncryptedSeed = await this._commonStorageAdapter.get(SeedVaultStorageKeys.ENCRYPTED_KEY);
+    if (prevStoredEncryptedSeed && prevStoredEncryptedSeed !== '' && !overwriteExistingSeed) {
```
victorkirov, line 56: Empty strings are falsey by default, no need for the extra check.

```suggestion
    if (prevStoredEncryptedSeed && !overwriteExistingSeed) {
```

seedVault/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    this._commonStorageAdapter.set(SeedVaultStorageKeys.SEED_VAULT_VERSION, this.VERSION.toString());
+    const salt = await this._cryptoUtilsAdapter.generateRandomBytes(32);
+    if (!salt) throw new Error('Salt not set');
+    const passwordHash = await this._cryptoUtilsAdapter.hash(password, salt);
+    if (!passwordHash) throw new Error('Password hash not set');
+    this._commonStorageAdapter.set(SeedVaultStorageKeys.PASSWORD_SALT, salt);
+    this._secureStorageAdapter.set(SeedVaultStorageKeys.PASSWORD_HASH, passwordHash);
+  };
+
+  storeSeed = async (seed: string, overwriteExistingSeed?: boolean) => {
```
victorkirov, line 52: I don't know if we should expose this like this 🤔 What about doing some thing like:

```
storeseed = async (seed: string) => this._storeseed(seed, false)

private _storeseed = async (seed: string, overwriteExistingSeed?: boolean) => {
...
```

PR #171, opened 2023-07-05
feat: build and publish test version on PRs

victorkirov:
This adds a publish step to PR builds (or if action is run manually on any branch) which publishes a test build to the package repository with a version `<current_package_version>-<commit_short_sha>`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   pull_request:
-    branches: [main, develop]
+    branches: [main]
 
 jobs:
   build:
--- a/.github/workflows/release-pull-request.yml
+++ b/.github/workflows/release-pull-request.yml
@@ -77,13 +77,12 @@ jobs:
         run: |
           curl https://api.github.com/repos/secretkeylabs/xverse-core/issues/170/comments \
           | jq ".[] | select(.user.login==\"github-actions[bot]\") | .id" \
-          | xargs -n1 -I %q curl \
+          | xargs -I %q curl \
             -L \
             -X DELETE \
-            -H "Accept: application/vnd.github+json" \
+            -H "Accept: application/vnd.github.v3+json" \
             -H "Authorization: token $GITHUB_TOKEN"\
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            https://api.github.com/repos/secretkeylabs/xverse-core/issues/comments/%q
+            https://api.github.com/repos/secretkeylabs/xverse-core/issues/comments/%q -vvv
       - name: Post test package PR comment
         if: ${{ github.event_name == 'pull_request' }}
         env:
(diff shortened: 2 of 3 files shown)


.github/workflows/release-pull-request.yml
```diff
@@ -0,0 +1,94 @@
...
+      - id: current-version
+        run: echo "version=$(npm pkg get version  | tr -d '"')" >> $GITHUB_OUTPUT
+      - id: tag-version
+        run: npm version --no-git-tag-version $CURRENT_VERSION-$SHA
+        env:
+          SHA: ${{ steps.git-commit.outputs.sha }}
+          CURRENT_VERSION: ${{ steps.current-version.outputs.version }}
+
+      - id: publish
+        run: npm publish
```
yknl, line 63: We should probably publish this to a "pr" tag so that it doesn't become the "latest" version

.github/workflows/release-pull-request.yml
```diff
@@ -0,0 +1,94 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - id: published-version
+        run: echo "version=$(npm pkg get version  | tr -d '"')" >> $GITHUB_OUTPUT
+      - run: echo published version $VERSION
+        env:
+          VERSION: ${{ steps.published-version.outputs.version }}
+
+      - name: Delete old bot comments
```
yknl, line 73: Is deleting older comments necessary? I think we can probably do without but open to be convinced. We can save ourselves from having a third-party action dependency.

victorkirov: Yeah, I was a bit torn on this one and having a third party dependency here. The comments get quite messy though because each subsequent commit to the branch would trigger a publish and a new comment.

I'll see if there isn't a way to delete comments via the API, like we do with the posting of the comment.

victorkirov: Cool, got it working via the API. Thanks for the motivation 😄

.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: Build
 
 on:
   pull_request:
-    branches: [main, develop]
+    branches: [main]
```
yknl, line 5: Should we have develop here? Since PRs are opened against develop.

victorkirov: The new workflow does this as one of its steps, so I took it out of here since it would be doing the same thing twice.

.github/workflows/release-pull-request.yml
```diff
@@ -0,0 +1,104 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: npm
+      - run: npm ci
+      # TODO: enable linting once ready
+      # - run: npm run lint
+      - run: npm test
```
victorkirov, line 29: @yknl this is the bit that the other workflow is doing

yknl (approved):

PR #173, opened 2023-07-12
feat: add the getSponsorInfo function to call /info endpoint

teebszet:
background:
- needed for swap transactions to be sponsored by a separate stacks-transaction-sponsor instance

changes:
- add a new function for getSponsorInfo to check if stacks-transaction-sponsor is active
- add types to both the stacks-transaction-sponsor endpoints
- improve error handling on stacks-transaction-sponsor endpoints
- allow supplying any host for the stacks-transaction-sponsor endpoints

Impact:
- limited to web extension
- added a new function getSponsorInfo
- updated the existing function sponsorTransaction (checked the codebases and this function is not used anywhere. xverse app uses a local version of the function)
- mostly backwards compatible. only change is the sponsorTransaction function now throws an error, but the function is so far not used anywhere

### DIFF
--- a/api/xverse.ts
+++ b/api/xverse.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosError, AxiosResponse } from 'axios';
+import axios, { AxiosResponse } from 'axios';
 import BigNumber from 'bignumber.js';
 import { API_TIMEOUT_MILLI, XVERSE_API_BASE_URL, XVERSE_SPONSOR_URL } from '../constant';
 import {
@@ -13,7 +13,6 @@ import {
   AppInfo,
   SponsorInfoResponse,
   SponsorTransactionResponse,
-  SponsorTransactionErrorResponse,
 } from 'types';
 import { StacksTransaction } from '@stacks/transactions';
 import { fetchBtcOrdinalsData } from './ordinals';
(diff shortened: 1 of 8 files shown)


api/xverse.ts
```diff
@@ -146,16 +149,34 @@ export async function getBinaceSignature(srcData: string): Promise<SignedUrlResp
...
-  return axios.post(sponsorUrl, data, { timeout: 45000 }).then((response) => {
-    return response.data.txid;
-  });
+  return axios
+    .post(sponsorUrl, data, { timeout: 45000 })
+    .then((response: AxiosResponse<SponsorTransactionResponse>) => {
+      return response.data.txid;
+    })
+    .catch(() => {
+      return null;
```
m-aboelenein, line 165: we should pass the error along to the apps from here to handle on the client side

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@secretkeylabs/xverse-core",
-  "version": "1.3.0",
+  "version": "1.4.0",
```
m-aboelenein, line 3: we usually update the package version in a separate pr

m-aboelenein (approved):

PR #175, opened 2023-07-14
fix: missing package.json dependency and type build errors

victorkirov:
resolves: ENG-2413

- added the missing dependency
- enabled typescript errors in webpack build and fixed them
- cleaned up unused variables

The missing dependency was used internally by the package that replaced it, at the same version we were using, so there were no breaking changes by removing it or adding it back in.


webpack.config.js
```diff
@@ -24,12 +25,14 @@ module.exports = {
       process: require.resolve('process/browser'),
       Buffer: ['buffer', 'Buffer'],
     }),
+    new ForkTsCheckerWebpackPlugin(),
```
m-aboelenein, line 28: we are using tsc to generate builds so this won't be affective in catching errors like the one in the issue

yknl (approved):

PR #182, opened 2023-07-20
feat: implement required calls to inscribe service for BRC20 transfer

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
This adds the calls to the Xverse inscribe api  to enable 1 step BRC20 transfers.

Issue Link: ENG-2391

# 🔄 Changes

Changes:
- Added calls to the inscribe service for brc20 transfer fee estimation, order creation and order execution.
- No breaking changes
 
Impact:
- With these endpoints we will be able to make 1-step brc20 transfers

# ✅ Review checklist

### DIFF
--- a/api/xverseInscribe.ts
+++ b/api/xverseInscribe.ts
@@ -1,0 +1,71 @@
+import axios from 'axios';
+
+import {
+  Brc20CostEstimateRequest,
+  Brc20CostEstimateResponse,
+  Brc20CreateOrderRequest,
+  Brc20CreateOrderResponse,
+  Brc20ExecuteOrderRequest,
+  Brc20ExecuteOrderResponse,
+  NetworkType,
+} from 'types';
+
+import { XVERSE_INSCRIBE_URL } from '../constant';
+
+const apiClient = axios.create({
+  baseURL: XVERSE_INSCRIBE_URL,
+});
+
+const getBrc20TransferFees = async (
+  tick: string,
+  amount: number,
+  revealAddress: string,
+  feeRate: number,
+): Promise<Brc20CostEstimateResponse> => {
+  const requestBody: Brc20CostEstimateRequest = {
+    operation: 'transfer',
+    tick,
+    amount,
+    revealAddress,
+    feeRate,
+  };
+  const response = await apiClient.post<Brc20CostEstimateResponse>('/v1/brc20/cost-estimate', requestBody);
+  return response.data;
+};
+
+const createBrc20TransferOrder = async (
+  tick: string,
+  amount: number,
+  revealAddress: string,
+  feeRate: number,
+  network: NetworkType,
+): Promise<Brc20CreateOrderResponse> => {
+  const requestBody: Brc20CreateOrderRequest = {
+    operation: 'transfer',
+    tick,
+    amount,
+    revealAddress,
+    feeRate,
+    network,
+  };
+  const response = await apiClient.post<Brc20CreateOrderResponse>('/v1/brc20/cost-estimate', requestBody);
+  return response.data;
+};
+
+const executeBrc20TransferOrder = async (
+  commitAddress: string,
+  commitTransactionHex: string,
+): Promise<Brc20ExecuteOrderResponse> => {
+  const requestBody: Brc20ExecuteOrderRequest = {
+    commitAddress,
+    commitTransactionHex,
+  };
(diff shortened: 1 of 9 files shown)


api/xverseInscribe.ts
```diff
@@ -0,0 +1,163 @@
...
+  const requestBody: Brc20CostEstimateRequest = {
+    operation: 'transfer',
+    tick,
+    amount,
+    revealAddress,
+    feeRate,
+    inscriptionValue,
+  };
+  const response = await apiClient.post<Brc20CostEstimateResponse>('/v1/brc20/cost-estimate', requestBody);
+  return response.data;
```
teebszet, line 35: no error handling?

transactions/brc20.ts
```diff
@@ -37,3 +60,306 @@ export const createBrc20TransferOrder = async (
...
+  });
+
+  const { commitAddress, commitValue } = await xverseInscribeApi.createBrc20MintOrder(
+    tick,
+    amount,
+    revealAddress,
+    feeRate,
+    network,
+    FINAL_SATS_VALUE,
+  );
```
teebszet, line 119: there's no error handling on the api method or here, so there's high potential for "destructuring of undefined" error here

transactions/brc20.ts
```diff
@@ -37,3 +60,306 @@ export const createBrc20TransferOrder = async (
...
+    new BigNumber(commitChainFees),
+    network,
+  );
+
+  const { revealTransactionId } = await xverseInscribeApi.executeBrc20Order(commitAddress, commitTransaction.hex);
+
+  return revealTransactionId;
+}
+
+export const brc20TransferEstimateFees = async (estimateProps: EstimateProps) => {
```
teebszet, line 154: add the return type here

transactions/brc20.ts
```diff
@@ -1,27 +1,50 @@
...
+import { generateSignedBtcTransaction, selectUtxosForSend, signNonOrdinalBtcSendTransaction } from './btc';
 
-const createTransferInscriptionContent = (token: string, amount: string) =>
-  ({
-    p: 'brc-20',
-    op: 'transfer',
-    tick: token,
-    amt: amount,
-  });
+const FINAL_SATS_VALUE = 1000;
```
teebszet, line 10: need explaining comment on this

transactions/brc20.ts
```diff
@@ -37,3 +60,306 @@ export const createBrc20TransferOrder = async (
...
+    accountIndex,
+    seedPhrase,
+    'Mainnet',
+    new BigNumber(transferFeeEstimate),
+  );
+
+  yield ExecuteTransferProgressCodes.Finalizing;
+
+  // we sleep here to give the reveal transaction time to propagate
+  await new Promise((resolve) => setTimeout(resolve, 500));
```
teebszet, line 345: we sleep even if the function paused at the last yield for some unknown time?

tests/transactions/brc20.test.ts
```diff
@@ -0,0 +1,422 @@
...
+  brc20MintExecute,
+  brc20TransferEstimateFees,
+  brc20TransferExecute,
+  ExecuteTransferProgressCodes,
+} from '../../transactions/brc20';
+
+vi.mock('../../api/xverseInscribe');
+vi.mock('../../api/esplora/esploraAPiProvider');
+vi.mock('../../transactions/btc');
+vi.mock('../../wallet');
```
teebszet, line 25: cool

transactions/brc20.ts
victorkirov: That shouldn't be the case. If the call throws an error, then the error will propagate down instead of returning `undefined`. We actually handle errors in quite a few places in core and return `null | undefined`, which we should probably remove. Returning undefined is very different to a call failing.

transactions/brc20.ts
victorkirov: Good point, will add 👍

transactions/brc20.ts
victorkirov: I think it makes sense, otherwise the caller would need to know to add a sleep after `yield ExecuteTransferProgressCodes.Finalizing`

transactions/brc20.ts
teebszet: ok, makes sense. I had it wrong, since I just looked up Axios Response and data will always be {}, or the call will throw.
so basically we don't want to return null or undefined from any of our api call methods

teebszet (approved):
thanks for answering my questions. lgtm

PR #191, opened 2023-07-26
fix: potential bug

victorkirov:
# 🔘 PR Type
- [X] Bugfix

Currently this is just a draft PR to see if the tests fail and get a review. I _think_ this is a bug 👀


yknl (approved):

yknl:
We need to test this to see the effects on final fee rate.

victorkirov:
Ok, great. I'll add a bunch of tests for that function and base it off some actual btc transactions to make sure they match 👍

victorkirov:
Fixed in https://github.com/secretkeylabs/xverse-core/pull/182

PR #195, opened 2023-07-27
feat: add logic for getting best UTXO combo for txn

victorkirov:
# 🔘 PR Type
- [X] Enhancement

# Screenshot
Txn made using this logic with fee rate set to 12: https://mempool.space/tx/d008c4788cf50a43a59d4c9e81a6642570bc9081bb951496dac2e0b7490597e6
image

NOTE: The requested fee rate is slightly higher when using this for transactions from a taproot address as the resulting transaction is smaller. We can expand on this in the future if we use this for ordinal transfers or other transactions from the taproot address.

# 📜 Background
Currently, we naively select which UTXOs to use for a transaction by picking UTXOs from a list ordered by block time (oldest to newest). This does not always end up with the most efficient transaction in terms of fees and best UTXO use. There is also an issue in how we calculate fees with the UTXO selection in that we add a change output during the calculation and UTXO selection, but the final transaction may not end up with enough SATS for change after the fees, so we end up with a transaction with 1 less output and, therefore, higher fees/fee rate.

The proposed solution is to create an algorithm which attempts to find the best UTXOs for a transaction by minimising the fees, while keeping the fee rate above the desired fee rate, while also either maximising the change or aiming for a zero change combination of UTXOs.

This means that the following are the desired outcomes (1 being more favourable), while keeping the fee at a minimum:
1. Having one or more input UTXOs which match the recipient sum+fees, as close as possible. This may result in a slightly higher fee rate than desired but would eliminate the change output (decreasing fees even though having a higher fee rate) and avoiding dust UTXOs.
2. Having the least number of inputs resulting in the biggest change output. This would use up UTXOs efficiently, while also ensuring that future transactions have a new UTXO with the highest value possible for more efficient future spend.

Issue Link: ENG-2497

# 🔄 Changes

Changes:
- Add a new function to select UTXOs for a transaction

Impact:
- This adds an improved way to select UTXOs for a transaction
- This method is currently unused in the extension and mobile app, so nothing to test yet

# ✅ Review checklist

### DIFF
--- a/tests/transactions/btc.fixtures.ts
+++ b/tests/transactions/btc.fixtures.ts
@@ -13,7 +13,13 @@ type SelectUtxoSendSuccessExpected = {
   fee: number;
 };
 
-type SelectUtxoSendSuccessFixture = [FixtureName, Recipient[], FeeRate, SelectUtxoSendSuccessExpected];
+type FixtureInput = {
+  recipients: Recipient[];
+  feeRate: FeeRate;
+  expected: SelectUtxoSendSuccessExpected;
+};
+
+type SelectUtxoSendSuccessFixture = [FixtureName, FixtureInput];
 
 export const recipientAddress1 = 'bc1pgkwmp9u9nel8c36a2t7jwkpq0hmlhmm8gm00kpdxdy864ew2l6zqw2l6vh';
 export const recipientAddress2 = 'bc1pyzfhlkq29sylwlv72ve52w8mn7hclefzhyay3dxh32r0322yx6uqajvr3y';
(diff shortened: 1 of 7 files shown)


tests/transactions/btc.fixtures.ts
```diff
@@ -0,0 +1,169 @@
...
+        address: recipientAddress1,
+        amountSats: new BigNumber(60000),
+      },
+    ],
+    10,
+    {
+      selectedUtxos: [utxo792k],
+      change: 730120,
+      fee: 1880,
+    },
```
teebszet, line 35: nit: i'd use named parameters and destructuring for these

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+  return undefined;
+}
+
+function selectOptimalUtxos(
+  recipients: Recipient[],
+  selectedUtxos: UTXO[],
+  availableUtxos: UTXO[],
+  changeAddress: string,
+  feeRate: number,
+  currentBestUtxoCount: number | undefined = undefined,
```
teebszet, line 629: prefer named params here. the last two params are number, number 👀

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+    const txSize = tx.vsize;
+    let sentSats = new BigNumber(0);
+
+    for (let i = 0; i < tx.outputsLength; i++) {
+      const output = tx.getOutput(i);
+      sentSats = sentSats.plus(new BigNumber((output.amount ?? 0n).toString()));
+    }
+
+    const change = sentSats.minus(recipientTotal);
+    const fee = new BigNumber(txSize).times(feeRate);
```
teebszet, line 571: seems like this could be a separate function if you need to scope the variables anyway

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+
+    const txSize = txNoChange.vsize;
+    let sentSats = new BigNumber(0);
+
+    for (let i = 0; i < txNoChange.outputsLength; i++) {
+      const output = txNoChange.getOutput(i);
+      sentSats = sentSats.plus(new BigNumber((output.amount ?? 0n).toString()));
+    }
+
+    const fee = inputSum.minus(sentSats);
```
teebszet, line 608: also this

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+  if (availableUtxos.length === 0 || wouldIncreaseFees) {
+    // we either have no more UTXOs to add or adding would make an existing outer selection worse, so we bail
+    return undefined;
+  }
+
+  // sort smallest to biggest so we can pop biggest from end
+  const sortedUtxos = [...availableUtxos];
+  sortedUtxos.sort((a, b) => a.value - b.value);
+
+  let bestSelectionData: TransactionUtxoSelectionMetadata | undefined = undefined;
```
teebszet, line 649:
```suggestion
  let bestSelectionData: TransactionUtxoSelectionMetadata?;
```

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+  // and skip if it does since it would be a worse selection
+  const wouldIncreaseFees = currentBestUtxoCount === selectedUtxos.length - 1;
+
+  if (availableUtxos.length === 0 || wouldIncreaseFees) {
+    // we either have no more UTXOs to add or adding would make an existing outer selection worse, so we bail
+    return undefined;
+  }
+
+  // sort smallest to biggest so we can pop biggest from end
+  const sortedUtxos = [...availableUtxos];
```
teebszet, line 646:
```suggestion
  const sortedUtxos: UTXO[] = [...availableUtxos];
```

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+  sortedUtxos.sort((a, b) => a.value - b.value);
+
+  let bestSelectionData: TransactionUtxoSelectionMetadata | undefined = undefined;
+  while (sortedUtxos.length > 0) {
+    const utxo = sortedUtxos.pop() as UTXO;
```
teebszet, line 651:
```suggestion
    const utxo = sortedUtxos.pop();
```
prefer typing on L646 to casting, as casting will just ignore type errors

transactions/btc.ts
teebszet, line 629:
```suggestion
  currentBestUtxoCount?: number,
```
the default param only gets assigned if the param is undefined, so defaulting to undefined is a noop

transactions/btc.ts
```diff
@@ -526,6 +527,201 @@ export async function getBtcFeesForNonOrdinalBtcSend(
...
+        selectedUtxos,
+        fee: fee.toNumber(),
+        feeRate: feeRate,
+        change: changeWithoutFee.toNumber(),
+      };
+    }
+  }
+
+  {
+    // try txn without change
```
teebszet, line 586: not sure I follow the logic compared to PR description:
> This means that the following are the desired outcomes (1 being more favourable), while keeping the fee at a minimum:
> 1. Having one or more input UTXOs which match the recipient sum+fees, as close as possible. This may result in a slightly higher fee rate than desired but would eliminate the change output (decreasing fees even though having a higher fee rate) and avoiding dust UTXOs.
> 2. Having the least number of inputs resulting in the biggest change output. This would use up UTXOs efficiently, while also ensuring that future transactions have a new UTXO with the highest value possible for more efficient future spend.

isn't this function trying 2. (with change) and then 1. (no change) ?

tests/transactions/btc.test.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, expect, it } from 'vitest';
+
+import BigNumber from 'bignumber.js';
+import { selectUtxosForSend } from '../../transactions/btc';
```
teebszet, line 4: 👏  for starting tests on this file.

I'd also consider unit testing each of the functions. smaller units for testing means other devs will debug faster

transactions/btc.ts
victorkirov: Will change to `let bestSelectionData: TransactionUtxoSelectionMetadata | undefined`, the initialisation to undefined is definitely redundant 🙈 . The question mark is for args though.

transactions/btc.ts
victorkirov: Good point, will extract 👍

transactions/btc.ts
victorkirov: They're actually pretty similar. Will try and pull into 1 function.

transactions/btc.ts
victorkirov: This type is already given by TS. There's no need to explicitly set it.

tests/transactions/btc.fixtures.ts
victorkirov: That's a great idea 👍

transactions/btc.ts
victorkirov: I did it again 🙈 Will fix 👍

transactions/btc.ts
victorkirov: This is on purpose. Maybe I should add comment 🤔 The sortedUtxos is a UTXO[] but the pop function could return undefined. Since the while loop checks that there are elements in the array, it is safe to cast it to UTXO. TS is fine with it, to completely bypass it you'd need to cast to unknown first `const utxo = sortedUtxos.pop() as unknown as UTXO`. 

Another option is to explicitly check that `utxo` is not undefined and throw an error if it is, though that should never happen due to the while loop check.

transactions/btc.ts
victorkirov: Oopsie 😅

tests/transactions/btc.test.ts
victorkirov: I hate exporting functions for the sake of testing 😅 They'll also end up in the module export since we're doing export * from; though they are definitely complex enough to require their own tests 🤔

victorkirov: Do you think it makes sense to pull these out into a file like brc20.utils.ts and then we can have tests for that file but not export the functions in this file? Then they won't get into the module export.

transactions/btc.ts
victorkirov: I added a comment in the code. Let me know if you think it's fine or if it should maybe be an explicit not undefined check 🙏

transactions/btc.ts
victorkirov: Yeah, this function is doing that. It will first ensure that the transaction cannot be made with change before trying with no change. The second part would always return a valid transaction if the first one would return one, except the bottom one would result in a potentially very large fee in that case.

So, all this function does is check if we can use the UTXOs for the transaction, and whether there would be change; it doesn't actually apply the logic in the PR description. That logic is applied in the `selectOptimalUtxos` method in this section:
```
/**
     * we want to:
     * - minimise fees
     * - have zero change or maximise change
     * - while staying above the selected fee rate
     **/
    if (!bestSelectionData) {
      bestSelectionData = nextSelectionData;
    } else if (bestSelectionData.fee > nextSelectionData.fee) {
      bestSelectionData = nextSelectionData;
    } else if (nextSelectionData.fee === bestSelectionData.fee) {
      if (bestSelectionData.change < nextSelectionData.change) {
        bestSelectionData = nextSelectionData;
      }
    } else {
      return bestSelectionData;
    }
```

A transaction without change would have a lower fee as it has one fewer outputs.

victorkirov: Let me know if there are any comments I can put in the code that would clear things up for future devs 🙏

tests/transactions/btc.test.ts
teebszet: ahh I see. yeah that makes sense. I see you did that already

transactions/btc.utils.ts
```diff
@@ -0,0 +1,187 @@
...
+  recipients: Recipient[];
+  selectedUtxos: UTXO[];
+  changeAddress: string;
+  feeRate: number;
+  privateKey: string;
+  recipientTotal: BigNumber;
+  withChange: boolean;
+}): TransactionUtxoSelectionMetadata | undefined {
+  const { privateKey, selectedUtxos, recipientTotal, recipients, changeAddress, feeRate, withChange } = props;
+  // try transaction with change
```
teebszet, line 22: I think the code moved and this comment no longer applies

transactions/btc.utils.ts
```diff
@@ -0,0 +1,187 @@
...
+    isValid = change.gt(BITCOIN_DUST_VALUE);
+  } else {
+    const inputSum = selectedUtxos.reduce<BigNumber>((sum, utxo) => sum.plus(utxo.value), new BigNumber(0));
+
+    change = new BigNumber(0);
+    fee = inputSum.minus(sentSats);
+    isValid = fee.div(txSize).gt(feeRate);
+  }
+
+  // check if there is change and if the change is greater than the vsize*feeRate + dust rate
```
teebszet, line 57: same here

teebszet (approved):

PR #196, opened 2023-07-31
chore: performance improvements for account derivation

fedeerbes:
# 🔘 PR Type

- [ ] Bugfix
- [X] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
This PR adds to new methods to optimize the performance of restoration of account from the seed phrase. It exposes one new method `restoreWalletWithAccountsMobile` which uses `getActiveAccountsFromRootNode` and `getWalletFromRootNode`. the idea is to get the root node once and use it where is necessary since the calculation is expensive in terms of CPU. 
It also adds the method `waitFrame`, a simple wrapper of requestAnimationFrame to let the CPU breathe between expensive operations 

Issue Link: #[ENG-2449](https://linear.app/xverseapp/issue/ENG-2449/performance-improvements-for-account-derivation)

# 🔄 Changes

Changes:
- Modify restoreWalletWithAccounts to generate the root node and uses a for loop instead of promise all to improve performance
- Create getActiveAccountsFromRootNode which receives the root node instead of generate it
- fetchActiveAccounts now uses getActiveAccountsFromRootNode to not introduce a breaking change
- Create getWalletFromRootNode which receives the root node instead of generate it
- walletFromSeedPhrase now uses getWalletFromRootNode to not introduce a breaking change
- Unify accountType definition
- None of this changes are breaking change

Impact:
- Improves performance for the mobile app when restoring accounts
- Impacts on mobile app and web extensions

# 🧪 Testing
We should test mobile and web extension in this PR and the following flows:
- Create account 
- Create wallet
- Restore wallet
- Restore accounts
- Change network on settings

### DIFF
--- a/account/index.ts
+++ b/account/index.ts
@@ -9,26 +9,22 @@ import {
   getOrCreateWalletConfig,
   updateWalletConfig,
 } from '../gaia';
-import {
-  Account,
-  BtcTransactionData,
-  NetworkType,
-  SettingsNetwork,
-  StxTransactionListData,
-} from '../types';
-import { bip32 } from '../utils/bip32';
-import { walletFromSeedPhrase } from '../wallet';
+import { Account, BtcTransactionData, NetworkType, SettingsNetwork, StxTransactionListData } from '../types';
+import { BIP32Interface, bip32 } from '../utils/bip32';
+import { getWalletFromRootNode, walletFromSeedPhrase } from '../wallet';
 import { GAIA_HUB_URL } from './../constant';
 
-export const fetchActiveAccounts = async (
-  mnemonic: string,
-  networkObject: StacksNetwork,
-  currentAccounts: Account[]
-) => {
+export const getActiveAccountsFromRootNode = async ({
+  networkObject,
+  currentAccounts,
+  rootNode,
+}: {
+  networkObject: StacksNetwork;
+  currentAccounts: Account[];
+  rootNode: BIP32Interface;
+}) => {
   const networkFetch = networkObject.fetchFn;
   const hubInfo = await getHubInfo(GAIA_HUB_URL, networkFetch);
-  const seed = await bip39.mnemonicToSeed(mnemonic);
-  const rootNode = bip32.fromSeed(Buffer.from(seed));
   const walletConfigKey = await deriveWalletConfigKey(rootNode);
   const currentGaiaConfig = connectToGaiaHubWithConfig({
     hubInfo,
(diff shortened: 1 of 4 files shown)


vite.config.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
m-aboelenein, line 1: why did we need to add this ?

wallet/index.ts
```diff
@@ -71,40 +70,46 @@ export async function newWallet(): Promise<BaseWallet> {
...
 
-  const master = bip32.fromSeed(seed);
-  const masterPubKey = master.publicKey.toString('hex');
   const stxPublicKey = publicKeyToString(getPublicKey(createStacksPrivateKey(privateKey)));
 
   // derive segwit btc address
+  await waitFrame();
   const btcChild = master.derivePath(getBitcoinDerivationPath({ index, network }));
+
+  await waitFrame();
```
m-aboelenein, line 101: i am not sure about this one did it make a notable difference in performance?

vite.config.ts
fedeerbes: cause it throwing an eslint error which doesn't make sense 

`'vitest' should be listed in the project's dependencies, not devDependencies.`

wallet/index.ts
fedeerbes: in mobile the difference is noticeable. we gain several frames between tasks with this

dhriaznov (approved):
LGTM

dhriaznov:
There's a git conflict to solve @fedeerbes

wallet/index.ts
fedeerbes: removed

vite.config.ts
fedeerbes: removed, not needed anymore

m-aboelenein (approved):

account/index.ts
```diff
@@ -43,11 +39,26 @@ export const fetchActiveAccounts = async (
   });
 };
 
+export const fetchActiveAccounts = async (
```
yknl, line 42: Does this function become obsolete after the seed vault PR https://github.com/secretkeylabs/xverse-core/pull/169?

fedeerbes: i have to take a look at that PR and get the context, maybe @m-aboelenein can have a quicker reply

m-aboelenein: this is not related to the seed vault yet it only handles storage of the seed phrase

PR #199, opened 2023-08-02
feat: add brc 20 transfer hook to core

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
We have hooks in the mobile app and the extension which have a lot of repeated code. This adds hooks to Core where we can build up a shared hook library. The first hook is the BRC-20 1 step transfer fees and execution hooks.

# 🔄 Changes

Changes:
- 2 new hooks for BRC-20 transfer fee estimates and execution
- Needs to be implemented in the extension and mobile app

# ✅ Review checklist


hooks/brc20/useBrc20TransferExecute.ts
```diff
@@ -0,0 +1,130 @@
...
+  accountIndex: number,
+  addressUtxos: UTXO[],
+  tick: string,
+  amount: number,
+  revealAddress: string,
+  changeAddress: string,
+  recipientAddress: string,
+  feeRate: number,
+  network: NetworkType,
+) => {
```
fedeerbes, line 40: consider using an object as param

hooks/brc20/useBrc20TransferExecute.ts
```diff
@@ -0,0 +1,130 @@
...
+  const [complete, setComplete] = useState(false);
+  const [transferTransactionId, setTransferTransactionId] = useState<string | undefined>();
+  const [progress, setProgress] = useState<ExecuteTransferProgressCodes | undefined>();
+  const [errorCode, setErrorCode] = useState<ErrorCode | undefined>();
+
+  const executeTransfer = () => {
+    setExecuted(true);
+  };
+
+  useEffect(() => {
```
fedeerbes, line 52: consider not using a useEffect here and just export method which is charge of executing, like executeTransfer. for example

```ts
const executeTransfer = () => {
const {errorCode} = validateParams()
if(errorCode) {
return setErrorCode(errorCode); 
}
try {
        const transferGenerator = await brc20TransferExecute({
...
}
```
that way you don't need executed state

hooks/brc20/useBrc20TransferExecute.ts
```diff
@@ -0,0 +1,130 @@
...
+        } else {
+          setErrorCode(ErrorCode.SERVER_ERROR);
+        }
+      }
+    };
+
+    runTransfer();
+  }, [executed, running]);
+
+  return { executeTransfer, transferTransactionId, complete, progress, error: errorCode };
```
fedeerbes, line 127: consider using something like `complete: !!transferTransactionId` here, so you can avoid the complete state

hooks/brc20/useBrc20TransferFees.ts
```diff
@@ -0,0 +1,88 @@
...
+  amount: number,
+  feeRate: number,
+  revealAddress: string,
+) => {
+  const [commitValue, setCommitValue] = useState<number | undefined>();
+  const [commitValueBreakdown, setCommitValueBreakdown] = useState<CommitValueBreakdown | undefined>();
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorCode, setErrorCode] = useState<ErrorCode | undefined>();
+
+  useEffect(() => {
```
fedeerbes, line 42: consider exposing a method to execute this code instead running it on the hook call. and since you don't need to update a progress state, you can just return an error or the fees values. no need to use state.

hooks/brc20/useBrc20TransferFees.ts
```diff
@@ -0,0 +1,88 @@
...
+    setIsLoading(true);
+    setErrorCode(undefined);
+
+    const runEstimate = async () => {
+      try {
+        const result = await brc20TransferEstimateFees({ addressUtxos, tick, amount, revealAddress, feeRate });
+        setCommitValue(result.commitValue);
+        setCommitValueBreakdown(result.valueBreakdown);
+      } catch (e) {
+        if (e.message === 'Not enough funds at selected fee rate') {
```
fedeerbes, line 72: proposal, maybe we can improve the error and use code and message. so we don't check against the message, that can change to provide more info later on.

hooks/brc20/useBrc20TransferExecute.ts
victorkirov: Oops, missed this one 🙈 Thanks, it's sorted

hooks/brc20/useBrc20TransferExecute.ts
victorkirov: 🙌 That's much better 😋

hooks/brc20/useBrc20TransferFees.ts
victorkirov: Yeah, that's a good idea. I'm gonna try make a custom error we can reuse in Core.

fedeerbes (approved):
LGTM

hooks/brc20/useBrc20TransferFees.ts
teebszet: @victorkirov what was the conclusion on this?

hooks/brc20/useBrc20TransferFees.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Estimates the fees for a BRC-20 1-step transfer
+ * @param addressUtxos - The UTXOs in the bitcoin address which will be used for payment
+ * @param tick - The 4 letter BRC-20 token name
+ * @param amount - The amount of the BRC-20 token to transfer
+ * @param feeRate - The desired fee rate for the transactions
+ * @param revealAddress - The address where the balance of the BRC-20 token lives. This is usually the ordinals address.
+ */
+const useBrc20TransferFees = (props: Props) => {
+  const { addressUtxos, tick, amount, feeRate, revealAddress } = props;
```
teebszet, line 32: @victorkirov why do we supply addressUtxos instead of finding it within the core function? like:
https://github.com/secretkeylabs/xverse-core/blob/2356a364559511b04ba331d390c738be1a419dce/transactions/btc.ts#L903

teebszet: and make feeRate optional, with default being the default feeRate?
like: https://github.com/secretkeylabs/xverse-core/blob/2356a364559511b04ba331d390c738be1a419dce/transactions/btc.ts#L784

hooks/brc20/useBrc20TransferFees.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  useEffect(() => {
+    const validationErrorCode = validateProps(props);
+
+    if (validationErrorCode) {
+      setErrorCode(validationErrorCode);
+
+      if (validationErrorCode !== ErrorCode.UTXOS_MISSING) {
+        setIsInitialised(true);
+      }
```
teebszet: what's the intended use for isInitialised?
is this because you're assuming there's an async fetch for addressUtxos, and you don't want to validate against the initial inputs?

I feel like this could be simplified by not calling runEstimate in useEffect, and instead exposing it as in the return. then you don't need to store all this state in the hook

(there's also an async fetch for feeRate)

utils/coreError.ts
```diff
@@ -0,0 +1,19 @@
...
+      this.stack += `\n\nCaused by: \n${causedBy.stack}`;
+    }
+  }
+
+  static isCoreError(e: Error): e is CoreError {
+    return e.name === 'CoreError';
+  }
+}
+
+export { CoreError };
```
teebszet, line 19: can we export this?

utils/coreError.ts
```diff
@@ -0,0 +1,19 @@
+class CoreError extends Error {
```
teebszet, line 1:
```suggestion
export class CoreError extends Error {
```

hooks/brc20/useBrc20TransferFees.ts
victorkirov: We could do both. The counter points for them which we need to consider are:
- For utxos, we usually have the useFees and useExecute hooks in one component. If they both get UTXOs individually, we would be calling the UTXO endpoint twice for the same data. This allows you to get UTXOs from outside the function and the collection can be reused in other parts of the calling component if needed.
- There's a similar argument for the feerate. Also, the component getting the default fee rate itself is a bit magical and side effecty and it's unclear what will happen if you don't specify a feerate.

I'm happy to build both in though if it would be better 👍

utils/coreError.ts
victorkirov: Yes, that is my bad 😅 Thanks for catching

hooks/brc20/useBrc20TransferFees.ts
teebszet: I wouldn't consider either as a blocker for this feature, but just want to align on the direction we're moving in.

because at the same time, we're refactoring this kind of logic to go into xverse-core: https://github.com/secretkeylabs/xverse-core/pull/201, to avoid mistakes and double fixing on mobile and extension

victorkirov: We should probably move it into core but move the useFeeRate and useAddressUTXOs hooks into core as well and then we can build cacheing into the hooks themselves to avoid multiple retrievals 🤔

teebszet (approved):

hooks/brc20/useBrc20TransferFees.ts
teebszet: i think this will be tackled in https://linear.app/xverseapp/project/consolidate-bitcoin-transaction-logic-da21d978a728/ENG ?

teebszet:
ready to merge this @victorkirov ?

hooks/brc20/useBrc20TransferExecute.ts
```diff
@@ -0,0 +1,137 @@
+import { useCallback, useState } from 'react';
+
+import { NetworkType, UTXO } from 'types';
```
fedeerbes, line 3: hey @victorkirov for some reason this was fixed before and now was reverted. right?

fedeerbes: btw eslint or ts should complain about this right?

victorkirov: Nope, that was from the first commit and hasn't changed since. What's the issue? The non-relative import?

victorkirov: I just changed it to relative

victorkirov: The way we've configured typescript for this project, it actually sees it as a valid import because we've set an alias:
image

I don't think we have anything at build time to process those aliases though, so not sure how it's supposed to work 🤔

fedeerbes: you're right. was another file that you fixed an import. i don't know why was causing errors on mobile if there is an alias

victorkirov: Did the import path change fix it?

victorkirov:
> ready to merge this @victorkirov ?

I think it's good to go now 👍

PR #203, opened 2023-08-08
feat: use xord for ordinal detection

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
We currently use Hiro API to detect ordinals in UTXOs, but that API is sometimes behind or incorrect. This switches to using Xord.

# 🔄 Changes

Changes:
We now use our own API for ordinal detection in UTXOs.

This will affect the ordinal recovery feature and ensuring that we don't spend ordinal UTXOs in BTC transactions.

Also added is the UTXO sort in the UTXO selection for a transaction. We now sort largest value to smallest and take from that list. This means that dust UTXOs would never be used for a transaction and it would result in a decrease in fees. 

# ✅ Review checklist


transactions/btc.ts
```diff
@@ -66,18 +66,9 @@ export function selectUnspentOutputs(
...
-      return 1;
-    } else if (b.status.block_time) {
-      return -1;
-    } else {
-      return a.value - b.value;
-    }
-  });
+  // Sort UTXOs based on value from largest to smallest
+  // This will give close to the optimal spend of UTXOs to minimise fees
+  unspentOutputs.sort((a, b) => b.value - a.value);
```
victorkirov, line 71: Note this change. It will select UTXOs for transactions in a way to minimise fees better. Hence all the changes in the tests.

DuskaT021:
@victorkirov there is a confict

victorkirov:
> @victorkirov there is a confict

Thanks @DuskaT021, sorted it out

api/ordinals.ts
```diff
@@ -47,48 +42,45 @@ const sortOrdinalsByConfirmationTime = (prev: BtcOrdinal, next: BtcOrdinal) => {
...
-          ordinals.push({
-            id: ordinal.data.id,
-            confirmationTime: utxo.status.block_time || 0,
-            utxo,
+    addressUTXOs
+      .filter((utxo) => utxo.status.confirmed) // we can only detect ordinals from confirmed utxos
+      .map(async (utxo: UTXO) => {
+        const ordinalContentUrl = `${XVERSE_INSCRIBE_URL}/v1/inscriptions/utxo/${utxo.txid}/${utxo.vout}`;
+
+        const ordinalIds = await axios.get<string[]>(ordinalContentUrl);
```
m-aboelenein, line 58: we should replace this block of code with the `getOrdinalIdFromUtxo` function

transactions/btc.ts
m-aboelenein: sorting by confirmation time allowed us to have a chance of `CPFP` to speed up transactions any chance we can do both here?

api/ordinals.ts
victorkirov: They do slightly different things. getOrdinalIdFromUtxo returns only one ID on a UTXO because that's what the original method did. This one gets IDs of all ordinals on a UTXO if there are multiple. We can rewrite getOrdinalIdFromUtxo to return multiple IDs as another issue where we'd have to update the extension and mobile app code appropriately.

transactions/btc.ts
victorkirov: Ah, didn't realise that. Will add it back if the values are the same.

yknl (approved):

transactions/btc.ts
```diff
@@ -66,17 +66,24 @@ export function selectUnspentOutputs(
     sumValue += pinnedOutput.value;
   }
 
-  // Sort UTXOs based on block time in ascending order
+  // Sort UTXOs based on value from largest to smallest
+  // This will give close to the optimal spend of UTXOs to minimise fees
   unspentOutputs.sort((a, b) => {
```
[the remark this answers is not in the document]
victorkirov: It's almost never beneficial to use the smallest one unless it is the perfect amount to not generate change. Using the smallest one would always incur higher fees if it doesn't covert the required amount completely. 

Using the smallest would also consume dust UTXOs, increasing the fees and not adding to the fee payments (the issue we're having now).

We'd have to think about the different strategies and how to expose them to the user. Maybe a power-user mode where they can pick pinned UTXOs for a txn or something.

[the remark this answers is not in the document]
yknl: yep makes sense

PR #212, opened 2023-08-11
Add STX support for Ledger accounts

dhriaznov:
# 🔘 PR Type
- [x] Enhancement

# 📜 Background
We would like to support STX account import, regular transactions and incoming txs signing as well.
We already have some code for STX support for Ledger accounts, but we should update and test it.

Issue Link: #[[ENG-2499](https://linear.app/xverseapp/issue/ENG-2499/add-stx-support-for-ledger-accounts)]
Context Link (if applicable):

# 🔄 Changes

Changes:
- Updated `importStacksAccountFromLedger` method logic
- Updated `signLedgerStxTransaction` method logic
- Added `StacksRecipient` type

Impact:
- This PR impacts the STX functionality for the ledger accounts (which weren't introduced to the users yet)

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


Imamah-Zafar (approved):

ledger/transaction.ts
```diff
@@ -155,12 +153,11 @@ export async function createNativeSegwitPsbt(
   return psbt;
 }
 
-export function addSignitureToStxTransaction(transaction: string | Buffer, signatureVRS: Buffer) {
-  const deserialzedTx = deserializeTransaction(transaction);
+export function addSignatureToStxTransaction(transaction: string | Buffer, signatureVRS: Buffer) {
```
teebszet, line 156: did we grep the codebases to see where this is being used, and open PRs to update?

one way to decouple the releases of xverse-core and consuming apps, would be to duplicate the function, and mark the old one with JSDocs: @deprecated comment. then we can be more flexible

ledger/transaction.ts
teebszet, line 1: was there any change to this file, other than correcting spelling mistakes? and prettier?

ledger/index.ts
```diff
@@ -521,41 +521,45 @@ export async function signSimpleBip322Message({
...
+    : await appStacks.getAddressAndPubKey(path, version);
 
   return { address, publicKey: publicKey.toString('hex') };
 }
 
 /**
  * This function is used to sign a Stacks transaction with the ledger
  * @param transport - the transport object with connected ledger device
- * @param transaction - the transaction to sign
+ * @param transactionBuffer - the transaction to sign
```
teebszet, line 550: was there a strong reason to change this interface and move the serialize call out of this function?

I noticed your web-extension PR didn't update confirmLedgerTransaction/index.ts L251 with the new param

and confirmStxTransaction/index.ts now does an extra .seralize() before calling this function

ledger/transaction.ts
dhriaznov: Agree, although in this particular case the function is ledger-specific and is only used in `ledger/index.ts` in core repo currently, so I believe that it should be safe to simply rename it in scope of this PR

ledger/transaction.ts
dhriaznov: Nope, those were the only changes here

ledger/index.ts
dhriaznov: Yeah, the serialization was added on the extension side because otherwise it threw an error:
Monosnap Xverse Wallet 2023-09-01 13-13-16

Currently, `confirmStxTransaction/index.ts` makes the tx serialization and passes it to `confirmLedgerTransaction/index.ts` screen ready to be passed to this core function in a Buffer format

dhriaznov:
Hey @teebszet, could you take a look once again please?

ledger/index.ts
teebszet: right. basically because we have to pass around a serialized transaction between screens anyway, so this removes the multiple actions to serialize (for location.state) -> deserialize (for this function) -> serialize (within this function)

is that right?

teebszet (approved):

ledger/index.ts
dhriaznov: Yeah, that's right

PR #217, opened 2023-08-17
chore: add optional parameters to getTransferTransactions and getConfirmedTransactions methods for offset and limit

fedeerbes:
# 🔘 PR Type

- [ ] Bugfix
- [X] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background
This PR adds missing offset and limit params to `getConfirmedTransactions` and `getTransferTransactions` methods.
I checked the usage of this methods and it seems like they're not used individually outside of `fetchStxAddressData` on mobile or extension, but i added the new params as optional so we maintain the backwards compatibility 

Issue Link: #[ENG-2637](https://linear.app/xverseapp/issue/ENG-2637/transaction-history-for-all-sip-10-tokens-stopped-displaying-on-mobile)

# 🔄 Changes

Changes:
- Add missing offset and limit params to `getConfirmedTransactions` and `getTransferTransactions` methods.

Impact:
- This PR improves the customization of the API calls made in the methods, so the clients can decide about limit and pagination
- In order to test this, you can use this beta version on mobile develop branch and inspect with proxyman if the API calls still work. Then you can go to `constants/constants.ts` on mobile and change the `PAGINATION_LIMIT` variable, you should see that the limit in the API calls changed to your value for reference i'll let the urls so is easier to test 
https://api.hiro.so/extended/v1/address/{{address}}/transactions
https://api.hiro.so/extended/v1/address/{{address}}/transactions_with_transfers

# ✅ Review checklist


api/stacks.ts
```diff
@@ -141,6 +155,8 @@ export async function fetchStxAddressData(
     getConfirmedTransactions({
       stxAddress,
       network,
+      offset: offset,
+      limit: paginationLimit,
```
teebszet, line 159: not sure how this is supposed to be used.. how will the caller of fetchStxAddressData know whether they've paginated through all of confirmedTransactions, mempoolTransactions and transferTransactions?

api/stacks.ts
```diff
@@ -107,10 +115,16 @@ export async function getMempoolTransactions({
...
   network: StacksNetwork,
+  offset?: number,
+  limit?: number,
 ): Promise<StxTransactionData[]> {
   const apiUrl = `${getNetworkURL(network)}/extended/v1/address/${stxAddress}/transactions_with_transfers`;
   const response = await axios.get<TransferTransactionsData>(apiUrl, {
     timeout: API_TIMEOUT_MILLI,
+    params: {
+      limit,
+      offset,
```
teebszet, line 126: does axios remove undefined params?

fedeerbes: it does

api/stacks.ts
fedeerbes: good question: the APIs return a "total" value available. so it can be calculated. but for i have seen we aren't not using pagination on mobile and extension, right? 
these changes just propagates the URL params to the inner methods

teebszet: does it solve the original bug?

I think we need to mark this function for refactor. 

- seems like getTransferTransactions returns a superset of getConfirmedTransactions, so not sure why we need to call both.
- the existing logic for adding transfer transactions to confirmed transactions doesn't account for pagination
- mempoolCount should be its total count including unpaginated pendingTransaction items

fedeerbes: the original issue on mobile won't be solved by these changes. cause was decided to not show the same tx that extension shows. It was decided to only increase the limit. 

the refactor requires additional discussion and work. it seems out of scope for this ticket, but i'm open to discuss if needed.

teebszet: @fedeerbes ok, does the other PR rely on this change? how about we put a comment in the code that this function has some known weird logic in case we see future bugs around pagination on this function

m-aboelenein:
is this pr still relevant ? @fedeerbes

fedeerbes:
> is this pr still relevant ? @fedeerbes

to my understanding yes, but was not prioritized. i'll bring up this in daily today

api/stacks.ts
fedeerbes: hey @teebszet i added a comment referencing this discussion for the future refactor https://github.com/secretkeylabs/xverse-core/pull/217/files#diff-deb0f3d60782723a978c7d4418ac3cc6d144b9f8682c382666d4ab6d7f335e33R55

teebszet (approved):

PR #241, opened 2023-09-15
feat: add a helper function for validate btc address is taproot

teebszet:
# 🔘 PR Type
- [x] Enhancement

# 📜 Background
Issue Link: https://linear.app/xverseapp/issue/ENG-2797/notice-for-an-non-taproot-address-when-sending-brc-20-tokens-web

required for PR: https://github.com/secretkeylabs/xverse-web-extension/pull/602

# 🔄 Changes

Changes:
- added helper function for validate btc address is taproot
- started a transactions/types.ts file for all exported types from transactions code

Impact:
- new function, to test with linked PR

# ✅ Review checklist

### DIFF
--- a/transactions/types.ts
+++ b/transactions/types.ts
@@ -1,7 +1,0 @@
-import BigNumber from 'bignumber.js';
-
-export type Brc20Recipient = {
-  address: string;
-  amountBrc20: BigNumber;
-  amountSats: BigNumber;
-};
--- a/transactions/types.ts
+++ b/transactions/types.ts
@@ -1,0 +1,7 @@
+import BigNumber from 'bignumber.js';
+
+export type Brc20Recipient = {
+  address: string;
+  amountBrc20: BigNumber;
+  amountSats: BigNumber;
+};
--- a/wallet/index.ts
+++ b/wallet/index.ts
@@ -322,5 +322,10 @@ export async function getStxAddressKeyChain(
 
 export { hashMessage };
 
-export const validateBtcAddressIsTaproot = (btcAddress: string): boolean =>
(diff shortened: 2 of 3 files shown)


transactions/types.ts
```diff
@@ -0,0 +1,7 @@
+import BigNumber from 'bignumber.js';
+
+export type Brc20Recipient = {
+  address: string;
+  amountBrc20: BigNumber;
+  amountSats: BigNumber;
+};
```
victorkirov, line 7: I don't think this needs to be in core. It can live in the extension repo since it's not used as an interface with core

wallet/index.ts
```diff
@@ -321,3 +321,11 @@ export async function getStxAddressKeyChain(
...
 
 export { hashMessage };
+
+export const validateBtcAddressIsTaproot = (btcAddress: string): boolean => {
+  try {
+    return getAddressInfo(btcAddress)?.type === AddressType.p2tr;
+  } catch {
+    return false;
+  }
+};
```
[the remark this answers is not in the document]
teebszet: one advantage of this approach is the consuming apps don't have to know anything about which library is being used in the implementation. 

I guess you could also return some custom enum types, but that seems a bit overkill?

transactions/types.ts
teebszet: agreed

teebszet: removed from core and added to extension on latest commits

wallet/index.ts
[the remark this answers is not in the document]
victorkirov: Good point 👍 Will mark as resolved

victorkirov (approved):
LGTM 🚢

PR #244, opened 2023-09-21
feat: add cancel token to fee estimate calls

victorkirov:
# 🔘 PR Type

- [x] Bugfix
- [x] Enhancement

# 📜 Background
When using the fee estimate hooks for BRC-20, if the value is changed too quickly, there is a chance that we fire off an estimation request twice, for each number typed by the user.

In the scenario that the first request finishes after the second, the UI would display the fee for the incorrect fee rate.

e.g. When typing a fee rate of 23, an estimation request is sent when the user types in '2' and then another when they type the '3' with a value of '23'. If the '23' call finishes first and the '2' call finishes second, the UI will show a fee rate of 23 but the fee estimates will be for '2'.

# 🔄 Changes

Changes:
- The fee estimate function now accepts an optional cancel token and the useEffect within the hook will cancel the api call when it is cleaned up

Impact:

- The BRC-20. 1 step hook will only show the fee estimates for the correct fee rate.

# Video
Before:

https://github.com/secretkeylabs/xverse-core/assets/3322667/de57c1cf-f06b-4e2f-9c2d-91e16b4376a0

After:

https://github.com/secretkeylabs/xverse-core/assets/3322667/860b858a-08fe-45e3-b3b5-43ec583a08ca

# ✅ Review checklist


yknl (approved):

yknl:
Tested and working

teebszet:
very nice!

PR #248, opened 2023-09-27
Mahmoud/eng 2739 utxo cache implementation

m-aboelenein:
# 🔘 PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background

Issue Link: #[issue_number]
Context Link (if applicable):

# 🔄 Changes

Changes:

- Which are breaking changes

Impact:

- What to test

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  CACHE_VERSION_STORAGE_KEY = 'utxoCacheVersion';
+
+  VERSION = 1;
+
+  constructor(config: UtxoCacheConfig) {
+    this._cacheStorageController = config.cacheStorageController;
+  }
+
+  _getCache = async (): Promise<UtxoCacheStruct> => {
```
victorkirov, line 32: Could you make these private? 🙏 The underscore is not super necessary if they are private, but I guess we should decide on that naming convention.
```suggestion
  private _getCache = async (): Promise<UtxoCacheStruct> => {
```

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+type UtxoCacheConfig = {
+  cacheStorageController: StorageAdapter;
+};
+
+export class UtxoCache {
+  private readonly _cacheStorageController: StorageAdapter;
+
+  CACHE_STORAGE_KEY = 'utxoCache';
+
+  CACHE_VERSION_STORAGE_KEY = 'utxoCacheVersion';
```
victorkirov, line 24: Could we rather tie the cache to the version directly to avoid an erroneous situation where we update the version and the app closes before the cache update is complete?

We could have a single storage key and the cache could look like:
```
type UtxoCache = {
  version: number;
  utxos: UtxoCacheStruct;
}
```

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
+import { getAddressUtxoOrdinalBundles, getUtxoOrdinalBundle } from './ordinals';
+
+export type UtxoStates = 'inscribed' | 'notInscribed' | 'unknown';
```
victorkirov, line 3: Maybe we should rename these to something more general now that we track rare sats 😅

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+  _getCache = async (): Promise<UtxoCacheStruct> => {
+    const cache = await this._cacheStorageController.get(this.CACHE_STORAGE_KEY);
+    if (cache) {
+      return JSON.parse(cache);
+    } else {
+      return {};
+    }
+  };
+
+  _getAddressUtxos = async (address: string) => getAddressUtxoOrdinalBundles(address, 0, 200);
```
victorkirov, line 41: Just a note that the API accepts a max limit of 60 and will truncate to 60 if a higher value is sent.

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  getUtxoState = async (utxoId: string): Promise<UtxoStates | null> => {
+    const cache = await this._getCache();
+    if (cache[utxoId]) {
+      return cache[utxoId];
+    }
+    const [txid, vout] = utxoId.split(':');
+    const utxo = await this._getUtxo(txid, parseInt(vout, 10));
+    const isInscribed = utxo.inscriptions.length > 0 ? 'inscribed' : 'notInscribed';
+    const isUnconfirmed = utxo.block_height === 0;
```
victorkirov, line 53: unconfirmed txns will actually have an undefined blockheight

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+    const utxoState = isUnconfirmed ? 'unknown' : isInscribed;
+    await this.setUtxoState(utxoId, utxoState);
```
victorkirov, line 55: There is a problem with the set/get logic in this method for 'unknown' transactions.

We either need to not tore 'unknown' state transactions in the cache, or we need to refetch those transactions on a get if we see that they have 'unknown' state.

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+    cache[utxoId] = state;
+    await this._cacheStorageController.set(this.CACHE_STORAGE_KEY, JSON.stringify(cache));
+  }
+
+  async removeUtxoState(utxoId: string): Promise<void> {
+    const cache = await this._getCache();
+    if (cache && utxoId in cache) {
+      delete cache[utxoId];
+    } else {
+      throw new Error('Utxo not found in cache');
```
victorkirov, line 73: It might be better to do a check if the utxoId is not in the cache then we just return rather than throw.

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    await this._cacheStorageController.set(this.CACHE_STORAGE_KEY, JSON.stringify(cache));
+  }
+
+  async clear(): Promise<void> {
+    await this._cacheStorageController.remove(this.CACHE_STORAGE_KEY);
+  }
+
+  async getVersion(): Promise<number> {
```
victorkirov, line 82: There is quite a mixture of class functions and class arrow function properties. Could you change them to either or please.

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
...
+      const isInscribed = utxo.inscriptions.length > 0 ? 'inscribed' : 'notInscribed';
+      const isUnconfirmed = utxo.block_height === 0;
+      acc[`${utxo.txid}:${utxo.vout}`] = isUnconfirmed ? 'unknown' : isInscribed;
+      return acc;
+    }, {} as { [key: string]: UtxoStates });
+    return utxosObject;
+  };
+
+  initCache = async (address: string): Promise<void> => {
+    await this._cacheStorageController.set(this.CACHE_VERSION_STORAGE_KEY, this.VERSION.toString());
```
victorkirov, line 105: We need to compare the version in storage to the current cache version and clear the cache if they don't match. This should be done throughout the class functions

api/utxoCache.ts
```diff
@@ -0,0 +1,109 @@
+import { getAddressUtxoOrdinalBundles, getUtxoOrdinalBundle } from './ordinals';
+
+export type UtxoStates = 'inscribed' | 'notInscribed' | 'unknown';
+
+export type UtxoCacheStruct = {
+  [utxoId: string]: UtxoStates;
```
victorkirov, line 6: I wonder if it won't be better and more useful if we make this:
```suggestion
  [utxoId: string]: UtxoOrdinalBundle;
```

That bundle info would actually be a lot more useful and it's not a huge payload. It will also never change in the long run (unless we do a cache version change for something like sattributes), so it's safe to store long term for confirmed/indexed UTXOs.

victorkirov (changes requested):
A few things to look into. I haven't gone through the tests yet, will do so after the stuff on the main code is resolved 😋

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+  ): Promise<void> => {
+    console.log('🚀 ~ file: utxoCache.ts:97 ~ UtxoCache ~ cachedUtxos:', cachedUtxos);
+    console.log('🚀 ~ file: utxoCache.ts:97 ~ UtxoCache ~ address:', address);
+    console.log('🚀 ~ file: utxoCache.ts:97 ~ UtxoCache ~ utxo:', utxo);
+    console.log('🚀 ~ file: utxoCache.ts:97 ~ UtxoCache ~ utxoId:', utxoId);
+    if (!cachedUtxos) {
+      cachedUtxos = {};
+    }
+    cachedUtxos[utxoId] = utxo;
+    await this._setCache(this.utxos, address);
```
victorkirov, line 106:
```suggestion
    await this._setCache(cachedUtxos, address);
```

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+type UtxoCacheConfig = {
+  cacheStorageController: StorageAdapter;
+};
+
+export class UtxoCache {
+  private readonly _cacheStorageController: StorageAdapter;
+
+  utxos: UtxoCacheStruct = {};
```
victorkirov, line 26: I don't think this is needed

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  VERSION = 1;
```
victorkirov, line 28: This should be readonly

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  constructor(config: UtxoCacheConfig) {
+    this._cacheStorageController = config.cacheStorageController;
+  }
+
+  private getAddressCacheStorageKey = (address: string) => `utxoCache-${address}`;
+
+  _getCache = async (address: string): Promise<UtxoCacheStorage> => {
```
victorkirov, line 36:
```suggestion
  private _getCache = async (address: string): Promise<UtxoCacheStorage> => {
```

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+    const cache = await this._cacheStorageController.get(this.getAddressCacheStorageKey(address));
+    if (cache) {
+      return JSON.parse(cache);
+    } else {
+      return {
+        version: -1,
+        utxos: {},
+      };
+    }
```
victorkirov, line 45: Might be good to wrap this in a try catch just in case

```suggestion
    if (cache) {
      try {
        return JSON.parse(cache);
      } catch(err) {
        // noop
      }
    } 
    return {
      version: -1,
      utxos: {},
    };
```

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  private _setCache = async (utxos: UtxoCacheStruct, address: string): Promise<void> => {
+    await this._cacheStorageController.set(
+      this.getAddressCacheStorageKey(address),
+      JSON.stringify({
+        version: this.VERSION,
+        utxos: utxos,
```
victorkirov, line 53:
```suggestion
        utxos,
```

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+      }),
+    );
+  };
+
+  _getAddressUtxos = async (address: string) => getAddressUtxoOrdinalBundles(address, 0, 60);
```
victorkirov, line 58:
```suggestion
  private _getAddressUtxos = async (address: string) => getAddressUtxoOrdinalBundles(address, 0, 60);
```

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  _getUtxo = async (txid: string, vout: number) => getUtxoOrdinalBundle(txid, vout);
```
victorkirov, line 60:
```suggestion
  private _getUtxo = async (txid: string, vout: number) => getUtxoOrdinalBundle(txid, vout);
```

api/utxoCache.ts
victorkirov: Maybe we should initiate the cache with all the UTXOs in the address instead of just the first 60

api/utxoCache.ts
```diff
@@ -0,0 +1,134 @@
...
+  getAllUtxos = async (btcAddress: string): Promise<UtxoCacheStruct> => {
+    const utxos = await this._getAddressUtxos(btcAddress);
+    const utxosObject: UtxoCacheStruct = utxos.results.reduce((acc, utxo) => {
+      acc[`${utxo.txid}:${utxo.vout}`] = utxo;
+      return acc;
+    }, {} as UtxoCacheStruct);
+    return utxosObject;
+  };
+
+  initCache = async (address: string): Promise<void> => {
```
victorkirov, line 127: I think there are too many functions exposed from this cache. The interface should be as simple as possible, so the only function exposed should be `getUtxoState` which returns the state for a single UTXO. The user of this class shouldn't need to manually check versions, initialise, and clear the cache.

That function should:
- try get the cache for the address
- if it doesn't exist or is wrong version, it re-initialise the cache, store it, and return the required UTXO
- if it does exist, load it and check if the UTXO is in the cache for the address
- if it is, return the state, if not, get that specific UTXO from the API, store it in the cache, and return the state

The getAllUtxos is a bit dangerous because the cache could be out of date and we might not have the latest UTXOs cached. Best to just allow a query to a single UTXO at a time.

victorkirov: It might actually be nice if we could initialise the cache in an in-memory mode with a wallet (so either 1 or 2 addresses) which it loads into memory. When we call the getUtxo method, it just works with the stuff in-memory and only stores those lookup objects into the cache when there are changes. If it's called directly with an address that's not in memory, it tries to load from storage for that address.

Alternatively, it could just keep the state in-memory for a max of 2 addresses and not need initialisation.

api/utxoCache.ts
```diff
@@ -0,0 +1,151 @@
...
+  initCache = async (address: string): Promise<UtxoCacheStruct | undefined> => {
+    const cacheUpdated = await this.isCacheUpdated(address);
+    if (!cacheUpdated) {
+      const utxos = await this.getAllUtxos(address);
+      await this._setCache(utxos, address);
+      return utxos;
+    }
+  };
+
+  setUtxo = async (
```
victorkirov, line 108: This should be private

api/utxoCache.ts
```diff
@@ -0,0 +1,151 @@
...
+    return utxo;
+  };
+
+  removeUtxo = async (utxoId: string, address: string): Promise<void | boolean> => {
+    const { utxos: cachedUtxos } = await this._getCache(address);
+    if (cachedUtxos && utxoId in cachedUtxos) {
+      delete cachedUtxos[utxoId];
+      await this._setCache(cachedUtxos, address);
+    }
+    return false;
```
victorkirov, line 145: Why do we return false here? If we return false on not found, maybe we can return true on found and deleted? Otherwise, just remove the return as it most likely won't be used anyways.

victorkirov (approved):
🚢

PR #258, opened 2023-10-18
Add stx nft collectible logic in core

Imamah-Zafar:
# 🔘 PR Type

- [ ] Bugfix
- [x] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background

Show user STX NFTs in collections 

Issue Link: #[[ENG-2847](https://linear.app/xverseapp/issue/ENG-2847/add-stx-nft-collectible-logic-in-core)]
Context Link (if applicable):

# 🔄 Changes

Changes:

- New functions introduced to fetch collection data for nfts 

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


stacksCollectible/index.ts
```diff
@@ -0,0 +1,145 @@
...
+  });
+
+  const requiredSortedCollectionsData = nftCollectionList?.slice(offset, offset + limit);
+
+  return {
+    offset,
+    limit,
+    total: nftCollectionList.length,
+    results: requiredSortedCollectionsData,
+  };
```
teebszet, line 144: so every time the user requests the next page, this function will refetch the entire dataset from the gamme API, sort it, and then return the requested page.

should the responses be cached between page requests?

teebszet: need @victorkirov's input here. because the extension/app will cache the returned pages, but not each request that core makes to gamma API. 

so current implementation would be something like:
1. user loads NFTs tab:
  -> core function paginates through all contracts
  -> core function calls collection endpoint for each contract
  -> core function sorts and returns page 1
  -> extension caches page 1
2. user clicks load more on NFTs tab:
  -> core function paginates through all contracts (again)
  -> core function calls collection endpoint for each contract (again)
  -> core function sorts and returns page 2
  -> extension caches page 2

stacksCollectible/index.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+  }
+
+  const nftCollectionList = Object.values(collectionRecord);
+
+  //sort according to total nft in a collection
+  nftCollectionList.sort((a, b) => {
+    //place bns collection at the bottom of nft list
+    if (a.collection_id === 'bns' || b.collection_id === 'bns') return -1;
```
victorkirov, line 133: You'll need to return 1 if a is bns and -1 if b is bns (or the other way around 🤔 )

stacksCollectible/index.ts
victorkirov: Yeah, this will definitely need caching within core. A few notes on this file:
- the call to `getAllNftContracts` should be cached for a reasonable TTL, even if it's just a plain in-memory cache. We don't need to get the full list each time.
- `getNftCollections` should be broken down into smaller functions. There are clear parts it can be broken down into to make the whole function easier to digest.
- `nftCollectionList` should be cached at the end and, when this function is called again, we just take a page out of the cached value to form the `requiredSortedCollectionsData` until the TTL expires.

victorkirov: We might not even need to cache `getAllNftContracts` if we're caching `nftCollectionList`

Imamah-Zafar: what's a reasonable TTL for storing this data? 2 hours or should it be something like a day

victorkirov: It will be an in-memory cache, so it doesn't really matter too much. Maybe 1 hour? The moment they close the extension/mobile app, the cache will be lost. The important bit for the cache is to not keep on hitting the API when they try to load more NFTs.

stacksCollectible/index.ts
Imamah-Zafar: we basically want to prioritize other collections over the 'bns' collection when sorting. So with the current implementation, if either a or b is a bns collection we assign -1 so it always comes at the end of the list. From my testing, all bns names do come at the end of the collection list. Is there some other reason why this lines needs to be updated 🤔

teebszet: @Imamah-Zafar the way the sort function works is it iterates through the original array by looking at two adjacent items (assigns them a and b). your compareFn supplied to sort() will tell it whether to swap a and b around or not.

compareFn should return 1 for swap a and b, or return -1 for do not swap a and b.
it means line 133 should return either 1 or -1 depending on which item was the bns item

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort
image

Imamah-Zafar: oh i see. thanks for this. I have updated the sorting

stacksCollectible/index.ts
```diff
@@ -0,0 +1,200 @@
...
+
+async function fetchBatchData(batch: Array<NonFungibleToken>, collectionRecord: Record<string, StacksCollectionData>) {
+  const nftDetailPromises: Array<Promise<NftDetailResponse>> = [];
+  const collectionDataPromises: Array<Promise<NftCollectionData | undefined>> = [];
+  const nftArray: Array<NonFungibleToken> = [];
+
+  batch.forEach((nft) => {
+    const principal: string[] = nft.asset_identifier.split('::');
+    const contractInfo: string[] = principal[0].split('.');
+    const contractId = principal[0];
```
teebszet, line 59:
```suggestion
    const principal: string[] = nft.asset_identifier?.split('::');
    const contractInfo: string[] = principal[0]?.split('.');
    const contractId = principal[0];
```

nit: I would usually be more defensive when dealing with external API responses like this

stacksCollectible/stacksCollectionCache.ts
```diff
@@ -0,0 +1,30 @@
+import { StacksCollectionData } from 'index';
+
+type RecordType = { userNftCollectionData: StacksCollectionData[]; ttl: number };
+//global value stored to be used for in-memory cache
+export const nftCollection: Record<string, RecordType> = {};
+
+export function setCacheValue(key: string, userNftCollectionData: StacksCollectionData[]) {
+  const DEFAULT_TTL = 1 * 60 * 60 * 1000; // ttl for 1 hour
+  // calculate ttl expiry time from current time
+  const currentTime = new Date().getTime();
```
teebszet, line 10:
```suggestion
  const currentTime = Date.now();
```

stacksCollectible/index.ts
teebszet: I think 1 hour might be too long for mobile. consider the following use cases:

1. user is navigating between tabs, and paginating stx NFTs
2. user has just bought a new NFT, and wants to see it

1 hour TTL will cover 1. but not be great for 2. on mobile. user would have to close the app and restart it to force a refetch

5mins TTL should still cover 1. but also 2.

stacksCollectible/stacksCollectionCache.ts
```diff
@@ -0,0 +1,30 @@
+import { StacksCollectionData } from 'index';
+
+type RecordType = { userNftCollectionData: StacksCollectionData[]; ttl: number };
+//global value stored to be used for in-memory cache
+export const nftCollection: Record<string, RecordType> = {};
+
+export function setCacheValue(key: string, userNftCollectionData: StacksCollectionData[]) {
+  const DEFAULT_TTL = 1 * 60 * 60 * 1000; // ttl for 1 hour
```
teebszet, line 8: I think 1 hour might be too long for mobile. consider the following use cases:

user is navigating between tabs, and paginating stx NFTs
user has just bought a new NFT, and wants to see it

1 hour TTL will cover 1. but not be great for 2. on mobile. user would have to close the app and restart it to force a refetch

5mins TTL should still cover 1. but also 2.

Imamah-Zafar: ok makes sense. then we should have a separate TTL for mobile (5min) and extension (1hr)  right ?

teebszet: there's no real downside of using 5mins TTL for both. extension is very easily closed and reopened, so it would be an edge case for the cache to last over 5mins anyway

stacksCollectible/stacksCollectionCache.ts
```diff
@@ -0,0 +1,30 @@
...
+  // store the data
+  nftCollection[key] = { userNftCollectionData, ttl };
+}
+
+export function getCacheValue(key: string) {
+  const val = nftCollection[key];
+  // if value doesn't exist
+  if (!val) return undefined;
+  // if ttl has expired
+  const currTime = new Date().getTime();
```
teebszet, line 22: nit:

```suggestion
  const currTime = Date.now();
```

teebszet (approved):

PR #263, opened 2023-10-27
fix: Import types via relative path and remove path aliases

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background

We had type aliases set up in tsconfig which were not being post-processed after a build. These were causing typing issues in our consuming applications.

# 🔄 Changes

Changes:

- Path aliases were removed from tsconfig.ts. 
- Only the "type" alias was used, so all imports from 'type' were changed to relative imports

Impact:

- This has no functional affect, only typing. It should fix some typing issues in the mobile and extension repos.

# ✅ Review checklist

### DIFF
--- a/api/gamma.ts
+++ b/api/gamma.ts
@@ -1,6 +1,6 @@
 import axios from 'axios';
-import { NftDetailResponse } from 'types';
 import { NFT_BASE_URI } from '../constant';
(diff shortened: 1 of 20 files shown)


teebszet (approved):
tested and it resolved my error! thanks @victorkirov

PR #264, opened 2023-10-31
chore: add a create release workflow

teebszet:
# 🔘 PR Type
- [x] CI related changes

# 📜 Background
Issue Link: https://linear.app/xverseapp/issue/ENG-2479/create-github-action-to-manage-xverse-core-release

Usage: 

use the new workflow for 'create-release-pr' which:
1. creates a release branch
2. bumps the version
3. pushes the branch and creates a PR
4. applies the label 'create release'
5. triggers a github draft release with change notes
6. comments back on the release PR with link to the draft release

Next steps:
either: continue to manually merge the PR, then publish the draft release, 
or: automate it so that when the PR is merged, it will trigger to set the draft release to latest and publish it

# 🔄 Changes
- new manual workflow: create a release PR, with options of: major, minor, patch
- new workflow: create github release, add release notes, comment on release PR


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,81 @@
+name: Create github release & release notes
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - main
+      - master
```
victorkirov, line 8: I think we can remove master

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,81 @@
...
+    types: [ labeled ]
```
victorkirov, line 9: How do we plan on running this workflow? I think the current configuration will potentially clash with the current PR beta publish workflow.

Maybe we could tie this down to PRs to main only? Then it won't clash with the other one that is only to develop.

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,81 @@
...
+  workflow_dispatch:
+
+jobs:
+  save-version:
+    if: github.event.label.name == 'create release'
```
victorkirov, line 14: Can you do a manual run (workflow_dispatch) with an event label? 🤔 If not, maybe we should remove the workflow_dispatch config above.

DuskaT021 (approved):

fedeerbes:
🚀🧉

PR #274, opened 2023-11-07
fix: add a full bns collection id for bns collectibles

teebszet:
# 🔘 PR Type
- [x] Bugfix
- [x] Enhancement

# 📜 Background
required for https://linear.app/xverseapp/issue/ENG-3145/add-bns-name-detail-page

# 🔄 Changes

Changes:
- use 10s timeout for nft holdings fetch
- supply bns contract id as collection id, so frontend can treat it the same to show contract id, name, etc.
- add some missing types

Impact:
- stx nft collectibles tab (bns collection and detail page)
- the 10s timeout for nft holdings means the nfts tab will show the API error message quicker (but should still retry with the react-query retries 3 times)

# ✅ Review checklist

### DIFF
--- a/api/stacks.ts
+++ b/api/stacks.ts
@@ -268,7 +268,7 @@ export async function getNftsData(
   const apiUrl = `${getNetworkURL(network)}/extended/v1/tokens/nft/holdings`;
 
   const response = await axios.get<NftEventsResponse>(apiUrl, {
-    timeout: 30000,
+    timeout: 10000,
     params: {
       principal: stxAddress,
       offset: offset,
--- a/stacksCollectible/index.ts
+++ b/stacksCollectible/index.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import { StacksNetwork } from '@stacks/network';
 import BigNumber from 'bignumber.js';
+import { BNS_CONTRACT_ID } from '../constant';
 import { NftCollectionData, NftDetailResponse, NftEventsResponse, NonFungibleToken } from 'types';
 import { getNftDetail, getNftsCollectionData } from '../api/gamma';
 import { getNftsData } from '../api/stacks';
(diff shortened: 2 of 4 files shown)


stacksCollectible/index.ts
```diff
@@ -59,9 +60,10 @@ async function fetchBatchData(batch: Array<NonFungibleToken>, collectionRecord:
     const contractId = principal[0];
 
     if (contractInfo[1] === 'bns') {
```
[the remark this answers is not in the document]
fedeerbes: and line 95 as well. if i'm not wrong

stacksCollectible/index.ts
```diff
@@ -59,9 +60,10 @@ async function fetchBatchData(batch: Array<NonFungibleToken>, collectionRecord:
...
-      //no further data required for BNS, arrange into collection
+      // no further data required for BNS, arrange into collection
+      // currently stacks only supports 1 bns name per address
```
fedeerbes, line 64: i know this was already merged but we shouldn't mutate a function param inside this fn [line 72](https://github.com/secretkeylabs/xverse-core/pull/274/files#diff-a02ab405a1cb56ad7eea17a73f1c8fb84e334bbcc8d5b332a44d21476c821a02R72) it can lead to undesired side effects imo

api/stacks.ts
```diff
@@ -268,7 +268,7 @@ export async function getNftsData(
   const apiUrl = `${getNetworkURL(network)}/extended/v1/tokens/nft/holdings`;
 
   const response = await axios.get<NftEventsResponse>(apiUrl, {
-    timeout: 30000,
+    timeout: 10000,
```
fedeerbes, line 271: out of curiosity, why the reduction of timeout? most of the APIs call uses API_TIMEOUT_MILLI

fedeerbes (approved):

stacksCollectible/index.ts
teebszet: yeah true

api/stacks.ts
teebszet: good point. this could do with more review, but the fetching of the STX collectibles dataset needs some optimisation in general.

things I noticed: 
- 30s timeout feels more like the UI is stuck rather than the data is loading
- react-query has a default to retry 3 times with exponential backoff before displaying error: https://tanstack.com/query/v4/docs/react/guides/query-retries. this could result in a 30s * 3 + 1000ms + 2000ms + 4000ms delay already before the user sees an error
- in addition to the fetch to this endpoint, each STX collection contract is then mapped to a promise.all to fetch collection info, which is another potential 30s * 3 + backoff delay

PR #282, opened 2023-11-13
fix: improve UTXO selection for spend

victorkirov:
# 🔘 PR Type

- [x] Bugfix
- [x] Enhancement

# 📜 Background

When selecting UTXOs to spend, we always take the biggest one as highest priority. This is problematic if a big UTXO is affected by a dust attack as it would be used over and over without being confirmed, potentially resulting in a chain of transactions with length of 25, the maximum allowed by Bitcoin.

If another transaction is executed and the last UTXO from that chain is the biggest, we would try to spend it again, resulting in a 400 from the Bitcoin node when trying to broadcast the transaction due to the long chain. This would effectively block your wallet from making transactions as that UTXO will always be picked first.

# 🔄 Changes

Changes:

- The logic to select which UTXOs to spend is changed, prioritising confirmed UTXOs primarily, and secondarily their value (biggest is higher priority)
- An additional step was added to further avoid spending dust UTXOs by ensuring that a UTXO added to a transaction increases the fee rate (if it decreases it, then it is dust at that fee rate)

Impact:

- Avoids the above described scenario with the long chain of txns
- Improves dust resiliency

# ✅ Review checklist

### DIFF
--- a/transactions/btc.ts
+++ b/transactions/btc.ts
@@ -57,6 +57,7 @@ export function selectUnspentOutputs(
   amountSats: BigNumber,
   unspentOutputs: Array<UTXO>,
   pinnedOutput?: UTXO,
+  feeRate?: number,
 ): Array<UTXO> {
   const inputs: Array<UTXO> = [];
   let sumValue = 0;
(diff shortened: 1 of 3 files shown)


transactions/btc.ts
```diff
@@ -66,27 +67,38 @@ export function selectUnspentOutputs(
...
+    } else if (!aConfirmed && bConfirmed) {
+      return 1;
     }
 
-    return 0;
+    return b.value - a.value;
   });
 
+  // this is slightly lower than the min vBytes of a segwit input
+  const minVBytesForInput = 70;
```
yknl, line 87: I believe inputs sizes can vary depending on the transaction type i.e. p2wpkh or p2tr

victorkirov: Yeah, that's true. This is just below the minimum of the smallest input for native segwit (which has the smallest vbytes per input). It's just a contingency check to ensure we never use extremely small UTXOs. The new logic looks at the actual size depending on the address.

yknl (approved):

PR #287, opened 2023-11-16
Featured Apps api utils

m-aboelenein:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background

Issue Link: #[[ENG-3261](https://linear.app/xverseapp/issue/ENG-3261/%5Bexpl-mob-1%5D-xverse-featured-apps-api)]
Context Link (if applicable):

# 🔄 Changes

Changes:

- Added the `getFeaturedDapps` method which fetches the list of featured dapps

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


jordankzf (approved):

types/api/xverse/appConfig.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface FeaturedDapp {
+  name: string;
+  url: string;
+  icon: string;
+  description: string;
+  order: number;
+  banner?: string;
+}
+
+export interface AppConfig {
```
victorkirov, line 19: nit: prefer types over interfaces

types/api/xverse/appConfig.ts
```diff
@@ -0,0 +1,21 @@
+export interface FeaturedDappOld {
+  name: string;
+  url: string;
+  image: string;
+  description: string;
+  order: number;
+  banner?: string;
+}
+
+export interface FeaturedDapp {
```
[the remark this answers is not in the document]
dhriaznov: There's more context on the API PR: https://github.com/secretkeylabs/xverse-api/pull/224#issuecomment-1857675358

api/xverse.ts
```diff
@@ -304,10 +307,16 @@ export async function getInscription(
   return response.data;
 }
 
-export async function getAppConfig(network: NetworkType) {
+export async function getAppConfig(network: NetworkType): Promise<AppConfig> {
   const appConfigUrl = `${XVERSE_API_BASE_URL(network)}/v1/app-config`;
-  const appConfig = await axios.get(appConfigUrl);
-  return appConfig;
+  const response = await axios.get(appConfigUrl);
+  return response.data.appConfig;
```
victorkirov, line 313: This looks like a breaking change 👀

api/xverse.ts
```diff
@@ -304,10 +307,16 @@ export async function getInscription(
...
+}
+
+export async function getFeaturedDapps(sectionsEnabled = false): Promise<(FeaturedDapp | FeaturedDappOld)[]> {
```
victorkirov, line 316: If `FeaturedDapp | FeaturedDappOld` is due to an api schema change, then we should be changing the end point to a v2 endpoint

types/api/xverse/appConfig.ts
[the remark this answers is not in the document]
victorkirov: That should've gone to a v2 endpoint :/ We don't want an API end point to return different schemas. Could we make the change in the API repo and change this to only have the new schema and point to the v2 endpoint?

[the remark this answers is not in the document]
victorkirov: It also looks like that PR went into the dev branch 👀 That is currently only hosted on render and the change wouldn't have made it into api-3.xveres.app. It needs to go into dev-k8s for that. That's our main branch on the API until we migrate api.xverse.app to k8s.

victorkirov (changes requested):
This needs some fixes

types/api/xverse/appConfig.ts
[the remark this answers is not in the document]
victorkirov: One last thing, looks like this type is actually not correct. The new response is a list of sections 👀

api/xverse.ts
dhriaznov: @m-aboelenein do we need this change?

types/api/xverse/appConfig.ts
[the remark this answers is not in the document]
dhriaznov: oh, I see.. @victorkirov should we make a new PR for the API with the fixes you mentioned and merge it into `dev` and then merge those changes into `dev-k8s` or should we revert that API PR and create a new one instead and merge it into `dev-k8s`?

api/xverse.ts
dhriaznov: Makes sense.. we won't need the `sectionsEnabled` param then, right?

api/xverse.ts
jordankzf: Removed for now

types/api/xverse/appConfig.ts
```diff
@@ -0,0 +1,18 @@
...
+  name: string;
+  url: string;
+  icon: string;
+  description: string;
+  order: number;
+  banner?: string;
+}
+
+export type DappSectionData = {
+    section: string;
```
dhriaznov, line 11: 4 space tabs here 😱

dhriaznov (approved):

PR #291, opened 2023-11-22
fix: Get UTXO ordinal data in batches

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background

When making payments from payments address, we only pick UTXOs that don't have inscriptions on them. We check the UTXOs by making a call to the backend one at a time. This leads to rate limits being hit.

# 🔄 Changes

Changes:

- We get all the inscriptions for the address in batches and map them to the UTXOs instead

Impact:

- This fixes the rate limiting issue and should speed up the loading of the confirmation pages for users with many UTXOs in their payments address

# ✅ Review checklist

### DIFF
--- a/api/ordinals.ts
+++ b/api/ordinals.ts
@@ -1,5 +1,6 @@
 import axios from 'axios';
 import BitcoinEsploraApiProvider from '../api/esplora/esploraAPiProvider';
+import XordApiProvider from '../api/ordinals/provider';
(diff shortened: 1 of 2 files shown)


teebszet:
cool. let's decouple this from the 0.23.0 release, and give more time to test

teebszet (approved):

PR #292, opened 2023-11-23
Enable RBF on our txns and add RBF logic

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background

We are adding RBF. This will add the required features for it.

# 🔄 Changes

Changes:

- Transaction inputs are now added with a sequence marking them as RBF enabled
- Added logic to generate the RBF replacement txn

Impact:

- We will need to add the RBF screens to the extension and mobile app

# ✅ Review checklist

### DIFF
--- a/transactions/rbf.ts
+++ b/transactions/rbf.ts
@@ -1,0 +1,100 @@
+import { RecommendedFeeResponse, Transaction } from '../types/api/esplora';
+
+// TODO: this is from txn consolidation logic
+// TODO: Use that instead once merged
+// TODO: Also, use that to sign these txns once merged
+const areByteArraysEqual = (a: undefined | Uint8Array, b: undefined | Uint8Array): boolean => {
+  if (!a || !b || a.length !== b.length) {
+    return false;
+  }
+
+  return a.every((v, i) => v === b[i]);
+};
+
+const isTransactionRbfEnabled = (transaction: Transaction) => {
+  if (transaction.status.confirmed) {
+    return false;
+  }
+
+  return transaction.vin.some((input) => input.sequence < 0xffffffff - 1);
+};
+
+const getRbfTransactionSummary = (transaction: Transaction) => {
+  const transactionVSize = transaction.weight / 4;
+
+  const currentTransactionInputTotals = transaction.vin.reduce((total, input) => total + input.prevout.value, 0);
+  const currentTransactionOutputTotals = transaction.vout.reduce((total, output) => total + output.value, 0);
+
+  const currentFee = currentTransactionInputTotals - currentTransactionOutputTotals;
+  const currentFeeRate = Math.ceil(currentFee / transactionVSize);
+
+  const minimumRbfFee = Math.ceil(transaction.fee + transactionVSize);
+  const minimumRbfFeeRate = Math.ceil(minimumRbfFee / transactionVSize);
+
+  return { currentFee, currentFeeRate, minimumRbfFee, minimumRbfFeeRate };
+};
+
+type TierFees = {
+  fee: number;
+  feeRate: number;
+};
+
+type RbfRecommendedFees = {
+  medium?: TierFees;
+  high?: TierFees;
+  higher?: TierFees;
+  highest?: TierFees;
+};
+
+const getRbfRecommendedFees = (transaction: Transaction, mempoolFees: RecommendedFeeResponse): RbfRecommendedFees => {
+  const { minimumRbfFeeRate } = getRbfTransactionSummary(transaction);
+
+  const { halfHourFee: medium, fastestFee: high } = mempoolFees;
+
+  if (minimumRbfFeeRate < medium) {
+    return {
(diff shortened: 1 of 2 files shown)


transactions/rbf.ts
```diff
@@ -0,0 +1,435 @@
...
+
+  private p2tr!: btc.P2TROut;
+
+  private network!: typeof btc.NETWORK;
+
+  private wallet!: RBFProps;
+
+  private paymentUtxos?: UTXO[];
+
+  constructor(transaction: Transaction, wallet: P) {
```
dhriaznov, line 100: We use the `BtcTransactionData` interface https://github.com/secretkeylabs/xverse-core/blob/a062f08934e8c8c2869f95d0f4bc6b92fbec3ff0/types/api/blockcypher/wallet.ts#L58 when fetching an array of txs or a single one on the extension, instead of the `Transaction` interface used in this file, it has a different `status` property and stuff. Should we change it here or adapt the extension to pass another object?

transactions/rbf.ts
```diff
@@ -0,0 +1,435 @@
...
+  return transaction.vin.some((input) => input.sequence < 0xffffffff - 1);
+};
+
+type RBFProps = {
+  btcAddress: string;
+  ordinalsAddress: string;
+  btcPublicKey: string;
+  ordinalsPublicKey: string;
+  seedVault: SeedVault;
+  addressIndex: number;
```
dhriaznov, line 53: Is this `addressIndex` the same property as `deviceAccountIndex` in our `Account` interface?https://github.com/secretkeylabs/xverse-core/blob/d1a9ef806d90b680dffaee8dbaa78430774cdcf7/types/account.ts#L3

transactions/rbf.ts
victorkirov: I think we may need to keep it as is. The BtcTransactionData doesn't seem to have `weight` which is used in the RBF logic.

transactions/rbf.ts
victorkirov: I think it's actually the `id` field on that interface.

I just checked in the Core library codebase and deviceAccountIndex doesn't seem to be used at all. Maybe it's something deprecated that we should clean up?

transactions/rbf.ts
victorkirov: I see we actually map the Transaction object to the BtcTransactionData object. I'll make the change to add the weight to BtcTransactionData and we'll be able to use it 👍

transactions/rbf.ts
dhriaznov: It's used for ledger-related logic, we pass it from web-extension so I think that on the core side it's used under some different param name

dhriaznov: Should we rename the `addressIndex` here to `id` or `accountId` to make it more obvious?

victorkirov: Yeah, that's a good idea. Will change it to addressIndex 👍

victorkirov: Hmmm, I'm not sure. This is what the derivation path function looks like
image

I think maybe we are using account and address id and index in an inconsistent way 🤔 

I guess accountId is probably the best option with the current naming. Will change it to that 👍

victorkirov: Though now we end up with this 😅 
image

I'll keep it as account index and we can add a tech debt ticket to make naming more conventional 👍

dhriaznov: I just checked the code and yeah, basically the `addressIndex` is for the `deviceAccountIndex` property, it's used to distinguish ledger accounts when user has more than one

dhriaznov: But it seems confusing, so we should probably rename it

victorkirov: Ah I see. Yeah, we'll need to rethink the naming with ledger and potentially other hw wallets around. I can see how this is actually confusing when we have the non-optional id field and the optional deviceAccountIndex field on the Account type.

I think you can see what makes the most sense in the arg type for RBF while you build with it and let me know what makes the most sense in terms of naming and structure and I'll put those changes in 👍

transactions/rbf.ts
```diff
@@ -0,0 +1,435 @@
...
+
+    return this.constructRecommendedFees(
+      'higher',
+      Math.ceil(minimumRbfFeeRate * 1.1),
+      'highest',
+      Math.ceil(minimumRbfFeeRate * 1.2),
+    );
+  };
+
+  getReplacementTransaction = async (options: O) => {
```
dhriaznov, line 420: The `options` param here requires to contain the `ledgerTransport` object, though it's not needed for regular account tx. Can we make it optional?
Screenshot 2023-11-24 at 21 22 26

dhriaznov: It looks like it should require to contain `feeRate` property instead

dhriaznov (approved):
Good job @victorkirov ✨

PR #293, opened 2023-11-23
Fix ordinal in btc address detection

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background

When sending an ordinal, we would detect if it is coming from the payments address and only sign with the payment address PK if that was the case. However, our detection logic would return true if there were any inscriptions in the payment address, not just the one we wanted to send.

This meant that if you were sending an ordinal and you happen to have another ordinal in your payments address, then we would sign the entire transaction with the payments PK, which would result in an invalid transaction.

# 🔄 Changes

Changes:

- We now detect if the UTXO is actually coming from the payments address or not

Impact:

- We need to update core in the extension and mobile and do the following test:
  - Ensure you have an inscription in your btc address
  - try and send an ordinal from your ordinals address
  - in the confirmation page, change the fee
  - try and send

The above should fail in the current version of the extension but succeed with this version of core.

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist

### DIFF
--- a/transactions/btc.ts
+++ b/transactions/btc.ts
@@ -807,8 +807,10 @@ export async function signOrdinalSendTransaction(
   // Make sure ordinal utxo is removed from utxo set used for fees
   // This can be true if ordinal utxo is from the payment address
 
+  const ordinalUtxoInPaymentAddress = unspentOutputs.some(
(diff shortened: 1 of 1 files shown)


dhriaznov (approved):
Good catch @victorkirov 🙌

PR #299, opened 2023-11-28
Add repetition args for inscriptions

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background

Allow for repeat inscription mints

# 🔄 Changes

Changes:

- Switch to using v2 endpoint for inscriptions
- Adds the repetitions field and uses the totalInscriptionValue field from the response

Impact:

- It's backwards compatible with the previous implementation, so it can be updated in place
- For repeat inscriptions, the caller will need to pass in the optional `repetitions` arg

# ✅ Review checklist


transactions/inscriptionMint.ts
```diff
@@ -244,6 +253,7 @@ export async function inscriptionMintExecute(executeProps: ExecuteProps): Promis
     availableUtxos: addressUtxos,
     feeRate,
     network,
+    useUnconfirmed: false,
```
victorkirov, line 256: Note this change. We are no longer allowing the use of unconfirmed UTXOs as inputs for inscription txns.

types/api/xverseInscribe/inscription.ts
```diff
@@ -35,6 +40,7 @@ export type InscriptionCreateOrderResponse = {
   commitValue: number;
   commitValueBreakdown: {
     inscriptionValue: number;
+    totalInscriptionValue: number;
```
teebszet, line 43: @victorkirov does this get passed to the hook return values?

teebszet:
can't verify with manual test yet...

types/api/xverseInscribe/inscription.ts
victorkirov: It was not. Good catch 👍  😄

PR #303, opened 2023-12-01
Don't use dust UTXOs in txns

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background

We were using confirmed Dust UTXOs as inputs for txns before unconfirmed large UTXOs (e.g. change).

Additionally, we weren't filtering out unconfirmed UTXOs when selecting them for inscription commit txns, so the fee estimate logic was using different logic to the final commit txn build logic causing inconsistencies.

# 🔄 Changes

Changes:

- When selecting UTXOs, if a fee rate has not been specified, we use the regular fee rate to filter out dust UTXOs

Impact:

- We would no longer use dust UTXOs for txns

# ✅ Review checklist

### DIFF
--- a/transactions/btc.ts
+++ b/transactions/btc.ts
@@ -16,10 +16,10 @@ const MINIMUM_CHANGE_OUTPUT_SATS = 1000;
 export const defaultFeeRate = {
   limits: {
     min: 5,
-    max: 10,
+    max: 50,
   },
-  regular: 5,
-  priority: 10,
+  regular: 12,
+  priority: 20,
 };
 
 export interface Recipient {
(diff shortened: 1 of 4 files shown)


transactions/btc.ts
```diff
@@ -55,9 +55,9 @@ export async function isCustomFeesAllowed(network: NetworkType, customFees: stri
 
 export function selectUnspentOutputs(
   amountSats: BigNumber,
+  feeRate: number,
```
teebszet, line 58: should we also refactor this function to use named args?

teebszet (approved):

PR #305, opened 2023-12-04
Use custom url for electrs api

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background

Currently, when changing the electrs URL in the apps, we don't use the custom URL for most functions. This ensures that all the places that use the URL have the api provider injected instead of being created in place, allowing the injection of a provider with a custom URL. 

# 🔄 Changes

Changes:

- All functions that created their own instance of the Esplora API Provider now accept it as an argument instead.

Impact:

- We will need to change the web extension and mobile apps to inject the provider in all the changed functions.

# ✅ Review checklist

### DIFF
--- a/ledger/psbt.ts
+++ b/ledger/psbt.ts
@@ -18,6 +18,7 @@ const areByteArraysEqual = (a?: Uint8Array, b?: Uint8Array): boolean => {
 const embellishNativeSegwitInputs = async (
   transaction: btc.Transaction,
   publicKey: string,
+  esploraProvider: EsploraProvider,
   network: NetworkType,
   addressIndex: number,
   masterFingerPrint: string,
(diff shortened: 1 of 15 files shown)


ledger/psbt.ts
```diff
@@ -153,7 +155,16 @@ export async function signLedgerPSBT({
...
-  if (await embellishNativeSegwitInputs(txn, nativeSegwitPubKey, network, addressIndex, masterFingerPrint)) {
+  if (
+    await embellishNativeSegwitInputs(
+      txn,
+      nativeSegwitPubKey,
+      esploraProvider,
+      network,
+      addressIndex,
+      masterFingerPrint,
+    )
```
victorkirov, line 166: There are a couple of functions like this that we should be moving to an object arg instead. I'd like to tackle that as a separate PR once txn consolidation is in and in use as a lot of the affected functions will be removed.

teebszet (approved):

PR #346, opened 2024-01-10
Refactor transaction logic for ease of use

victorkirov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background
Minor changes to make the consolidated logic nicer to consume

# 🔄 Changes

Changes:

- Mostly new types added and some renaming

Impact:

- Nothing is using this logic as yet, so no changes in extension/mobile needed

# ✅ Review checklist


m-aboelenein (approved):

transactions/bitcoin/context.ts
```diff
@@ -518,6 +579,8 @@ export class LedgerP2trAddressContext extends P2trAddressContext {
       throw new Error('Transport is required for Ledger signing');
     }
 
+    super.prepareInputs(transaction, { ...options, addNonWitnessUtxo: true });
```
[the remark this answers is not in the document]
victorkirov: I don't think that would work. If we use `this` here, the function would just call itself instead of the function on the base class.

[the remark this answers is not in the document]
m-aboelenein: can we make this a util function then instead of a base class method?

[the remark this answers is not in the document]
victorkirov: The base classes could have different implementations depending on the type of address, so it should be in the base class.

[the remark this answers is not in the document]
victorkirov: Actually, I see what you mean with this one. a util function makes sense here. Will refactor 🙌

currency/index.ts
```diff
@@ -15,20 +15,24 @@ const getStxFiatEquivalent = (stxAmount: BigNumber, stxBtcRate: BigNumber, btcFi
 const getBtcFiatEquivalent = (btcAmount: BigNumber, btcFiatRate: BigNumber): BigNumber =>
   satsToBtc(btcAmount).multipliedBy(btcFiatRate);
 
+const getFiatBtcEquivalent = (fiatAmount: BigNumber, btcFiatRate: BigNumber): BigNumber =>
+  new BigNumber(fiatAmount.dividedBy(btcFiatRate).toFixed(8));
```
teebszet, line 19: fiatAmount is already a BigNumber, and do we need toFixed(8) here, if we're also returning a BigNumber?

```suggestion
  fiatAmount.dividedBy(btcFiatRate);
```

teebszet: also should we deprecate the function getBtcEquivalent below? or did you specifically want a new function that did toFixed(8) as well?

transactions/bitcoin/index.ts
```diff
@@ -32,22 +33,55 @@ export type {
...
   TransactionScriptOutput,
+  TransactionSummary,
 };
 
 /**
  * send max bitcoin
  */
-export const sendMaxBtc = async (context: TransactionContext, toAddress: string, feeRate: number) => {
-  const paymentUtxos = await context.paymentAddress.getUtxos();
+export const sendMaxBtc = async (context: TransactionContext, toAddress: string, feeRate: number, skipDust = true) => {
```
teebszet, line 42: could we get unit tests on this? or do they exist already?

currency/index.ts
victorkirov: Yeah, we do. 

e.g. a btc is worth 3USD and I want to send 1USD
1/3=0.333333333333333...
If we had to test it, we'd get:
```
> a=new BigNumber('1')
BigNumber { s: 1, e: 0, c: [ 1 ] }
> b=a.div('3')
BigNumber { s: 1, e: -1, c: [ 33333333333333, 33333300000000 ] }
> b.toString()
'0.33333333333333333333'
```

0.33333333333333333333 is not a valid btc value because a bitcoin can only be divided into 100 million sats, so we need to truncate it to 8 digits after the decimal.

victorkirov: Ooops, I didn't see the other function 😅 I think the name of the new one makes more sense for consistency with the other function names. wdyt?

transactions/bitcoin/context.ts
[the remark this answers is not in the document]
victorkirov: Extracted it to a protected function in the base class

transactions/bitcoin/index.ts
victorkirov: I actually need to add tests for this entire file 👍 Thanks for catching

currency/index.ts
teebszet: yeah agree on the naming.

> 0.33333333333333333333 is not a valid btc value because a bitcoin can only be divided into 100 million sats, so we need to truncate it to 8 digits after the decimal.

I see, OK

teebszet:
I wonder if it's worth effort to make this PR backwards compatible? I remember our discussion with @fedeerbes around reducing breaking changes for when mobile is forced to bump core for certain functionality, but then required to refactor/fix for breakages

victorkirov:
> I wonder if it's worth effort to make this PR backwards compatible? I remember our discussion with @fedeerbes around reducing breaking changes for when mobile is forced to bump core for certain functionality, but then required to refactor/fix for breakages

This is backwards compatible, purely from the fact that this code is not used in mobile yet 😋

victorkirov:
@m-aboelenein @teebszet ready for final review 🙏

teebszet:
> This is backwards compatible, purely from the fact that this code is not used in mobile yet 😋

does that mean we can use a minor version bump?

teebszet (approved):

victorkirov:
> > This is backwards compatible, purely from the fact that this code is not used in mobile yet 😋
> 
> cool. still technically a major version bump right?

Yup, there are some interface changes on things that weren't used before.

PR #348, opened 2024-01-15
Update the RbfTransaction class constructor

dhriaznov:
# 🔘 PR Type

- [x] Enhancement

# 📜 Background

We use the `getSeedPhrase` function to sign the transaction, so there is no need to ask for it in the RbfTransaction constructor, it would make more sense to pass it to the signing function. Also, this change would enable a better UX for the mobile app since we won't need to ask the user to enter a password when they just enter the Speed Up tx screen, instead, they'll have to enter the password only when they sign the RBF tx.

Issue Link: #[[ENG-3379](https://linear.app/xverseapp/issue/ENG-3379/make-ui-for-rbf-in-mobile-app)]
Context Link (if applicable):

# 🔄 Changes

Changes:

- The breaking change here is that the `getSeedPhrase` function is now being passed to the `getReplacementTransaction` method of the `RbfTransaction` class instead of its' constructor.

Impact:

- This change impacts the RBF functionality for both mobile and web-extension, so we should adjust the code there.

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist

### DIFF
--- a/transactions/rbf.ts
+++ b/transactions/rbf.ts
@@ -283,7 +283,7 @@ class RbfTransaction {
 
   private signTxSoftware = async (transaction: btc.Transaction, options?: CompileOptions): Promise<btc.Transaction> => {
     if (!options?.getSeedPhrase) {
-      throw new Error('Options are required for non-dummy transactions');
+      throw new Error('getSeedPhrase option is required to sign the transaction');
     }
 
     const tx = btc.Transaction.fromPSBT(transaction.toPSBT(0));
--- a/transactions/rbf.ts
+++ b/transactions/rbf.ts
@@ -106,7 +106,6 @@ export type RBFProps = {
   ordinalsAddress: string;
   btcPublicKey: string;
   ordinalsPublicKey: string;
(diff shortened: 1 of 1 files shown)


transactions/rbf.ts
```diff
@@ -274,16 +274,20 @@ class RbfTransaction {
...
+    const seedPhrase = await options.getSeedPhrase();
     const seed = await bip39.mnemonicToSeed(seedPhrase);
     return bip32.fromSeed(seed);
   };
 
-  private signTxSoftware = async (transaction: btc.Transaction): Promise<btc.Transaction> => {
+  private signTxSoftware = async (transaction: btc.Transaction, options?: CompileOptions): Promise<btc.Transaction> => {
+    if (!options?.getSeedPhrase) {
+      throw new Error('Options are required for non-dummy transactions');
+    }
```
fedeerbes, line 287: question: should we change the message to "getSeedPhrase option is required" here? since it's the single prop the code is validating

fedeerbes (approved):

PR #353, opened 2024-01-19
ENG-3615: getOrdinalsFtBalance Improvements

jordankzf:
# 🔘 PR Type

- Extends FungibleToken and Coin from BaseToken (non-breaking)
- ~Removes forced `visible: true` from getOrdinalsFtBalance (possibly breaking)~
- Returns ordinal ticker as principal (non-breaking)
- Enforce upper case ordinal ticker (non-breaking)

- [ ] Bugfix
- [x] Enhancement
- [x] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background

Issue Link: #[issue_number]
Context Link (if applicable):

# 🔄 Changes

Changes:

- Which are breaking changes

Impact:

~If a downstream app that uses getOrdinalsFtBalance previously assumes that visible is set to true, it will now have to manually initialize it as true.~

- What to test

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist


api/ordinals.ts
```diff
@@ -149,12 +149,14 @@ export async function getOrdinalsFtBalance(network: NetworkType, address: string
             balance: responseToken.overallBalance,
             total_sent: '0',
             total_received: '0',
-            principal: '',
+            principal: responseToken.ticker.toUpperCase(),
```
abdulhaseeb4239, line 152: why set the ticker as principal?

api/ordinals.ts
```diff
@@ -149,12 +149,14 @@ export async function getOrdinalsFtBalance(network: NetworkType, address: string
...
             assetName: '',
-            ticker: responseToken.ticker,
+            ticker: responseToken.ticker.toUpperCase(),
             decimals: 0,
             image: '',
-            visible: true,
+            // explicit undefined = unknown initial state,
+            // consumers may then set it to true/false depending on the user's settings
+            visible: undefined,
```
abdulhaseeb4239, line 159: we want to display the tokens in the dashboard by default that have balance in the wallet

api/ordinals.ts
jordankzf: `principal` represents a unique identifier for a given fungible token. For SIP-10 tokens, the `contract` is unique. For BRC-20, there is no contract, but the `ticker` is known to be unique.

api/ordinals.ts
jordankzf: The following snippet is from a downstream consumer, our web extension:
```
const useBtcCoinBalance = () => {
  const dispatch = useDispatch();
  const { ordinalsAddress, network, brcCoinsList } = useWalletSelector();

  const fetchBrcCoinsBalances = async () => {
    try {
      const list = await getOrdinalsFtBalance(network.type, ordinalsAddress);

      const mergedList = list.map((newToken) => {
        const existingToken = brcCoinsList?.find((et) => et.ticker === newToken.ticker);

        return {
          ...newToken,
          // The old code enforces the ticker as upper case in the store, even though it returns the lower case ticker
          // Removing this since BRC-20 tokens do not have contracts, and rely on the case-insensitive ticker as a unique identifier
          // (it's case-sensitive within our code, because it's a string)
          // ticker: newToken.ticker?.toUpperCase(),
          // The `visible` property from `xverse-core` defaults to true.
          // We override `visible` to ensure that the existing state is preserved.
          ...(existingToken ? { visible: existingToken.visible } : {}),
        };
      });
      dispatch(setBrcCoinsDataAction(mergedList));
      return mergedList;
    } catch (e: any) {
      return Promise.reject(e);
    }
  };
```
`xverse-core` in its current state would mean that the extension has to enforce upper case for the ticker to ensure a consistent value throughout its codebase, as well as ignore the returned `visible` attribute to prevent overriding the latest local state.

I'm proposing enforcing upper case ticker (and derivative principal) upstream here at core, to avoid hacky solutions in downstream consumers.

This change would mean that the xverse app would need to initialize `visible` as true if there is no known local state. Do you think this is okay?

jordankzf: Ideally, this is what I would like to achieve:
```
  const fetchBrcCoinsBalances = async () => {
    try {
      const list = await getOrdinalsFtBalance(network.type, ordinalsAddress);

      const mergedList = list.map((newToken) => {
        const existingToken = brcCoinsList?.find((et) => et.ticker === newToken.ticker);

        return {
          ...existingToken,
          ...newToken,
        };
      });
      dispatch(setBrcCoinsDataAction(mergedList));
      return mergedList;
    } catch (e: any) {
      return Promise.reject(e);
    }
  };
```
Where we can simply combine `existingToken` & `newToken`, with the `newToken` attributes taking priority over the "stale" data. This way, we don't have to explicitly cherry-pick changes to prevent overriding local states.

abdulhaseeb4239: makes sense!

abdulhaseeb4239 (approved):

PR #354, opened 2024-01-19
[ENG-3617] Update the rbf logic

dhriaznov:
# 🔘 PR Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] Build related changes
- [ ] CI related changes

# 📜 Background

Issue Link: #[issue_number]
Context Link (if applicable):

# 🔄 Changes

Changes:

- Which are breaking changes

Impact:

- What to test

# 🖼 Screenshot / 📹 Video

# ✅ Review checklist

### DIFF
--- a/transactions/rbf.ts
+++ b/transactions/rbf.ts
@@ -112,7 +112,7 @@ export type RBFProps = {
   esploraProvider: EsploraProvider;
 };
 
-type TierFees = {
+export type TierFees = {
   enoughFunds: boolean;
   fee?: number;
   feeRate: number;
@@ -124,7 +124,7 @@ type CompileOptions = {
   getSeedPhrase: () => string | Promise<string>;
 };
 
-type RbfRecommendedFees = {
+export type RbfRecommendedFees = {
   medium?: TierFees;
   high?: TierFees;
   higher?: TierFees;


hooks/useRbfTransactionData/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+  if (!Number.isSafeInteger(minimumFee)) {
+    // round up the fee to the nearest integer
+    minimumFee = Math.ceil(minimumFee);
+  }
+
+  if (fee.lt(BigNumber(mediumFee))) {
+    feePresets = constructRecommendedFees('medium', mediumFee, 'high', highFee, stxAvailableBalance);
+  } else {
+    feePresets = constructRecommendedFees('higher', higherFee, 'highest', highestFee, stxAvailableBalance);
+  }
```
teebszet, line 102: this is the business logic here that we need to unit test in helpers.test.ts

the unit tests should serve two purposes: 
- formalise the business logic intention, uncovering edge cases in the process
- prevent regressions in future when we need to alter the business logic, or extend functionality

how you do it will be down to code style, but if it were me, I would extract L78-102 into a separate pure function. then you won't need to mock the functions with side effects like estimateTransaction, making the test file lighter and less coupled to code which we are not trying to test here

hooks/useRbfTransactionData/helpers.test.ts
teebszet, line 1: thank you for writing unit tests! 👏 

small nitpick that in xverse-core there is a convention to put test files in tests/ which is necessary in case we don't want to bundle the test files in with the published npm package

hooks/useRbfTransactionData/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        appInfo,
+        stxAvailableBalance,
+      );
+      setRbfData(calculatedData);
+    } catch (err: any) {
+      setErrorCode('SOMETHING_WENT_WRONG');
+    } finally {
+      setIsLoading(false);
+    }
+  }, [transaction, btcNetwork, stacksNetwork, appInfo, stxAvailableBalance]);
```
teebszet, line 49: I still think the stx rbf belongs in its own hook 🙂

transactions/stx.ts
```diff
@@ -454,4 +457,21 @@ export async function generateContractDeployTransaction(options: {
   }
 }
 
+export async function getLatestNonce(stxAddress: string, network: SettingsNetwork): Promise<LatestNonceResponse> {
+  const baseUrl = network?.address;
+  const apiUrl = `${baseUrl}/extended/v1/address/${stxAddress}/nonces`;
```
teebszet, line 462: it will probably throw an axios error even with the optional chaining operator `?.` since the apiUrl will be malformed

maybe we just rely on the type checking, since network is a required param
```suggestion
  const baseUrl = network.address;
  const apiUrl = `${baseUrl}/extended/v1/address/${stxAddress}/nonces`;
```

hooks/useRbfTransactionData/helpers.test.ts
dhriaznov: oh okay, moving it to `tests/hooks/useRbfTransactionData/helpers.test.ts`

hooks/useRbfTransactionData/helpers.ts
dhriaznov: yeah, in fact this exact thing is what is currently covered in `helpers.test.ts` in two assertions, `should return the fee estimation fetched from the Stacks function` and `should return the fee estimation based on the current fee`. do you think we should write more tests for this logic?

fedeerbes, line 102: nit: i always found better to define const to avoid any possible mistake while reassigning the var also can save you from if statements 

```ts
const shouldCapFee = appInfo?.thresholdHighStacksFee && high.fee > appInfo.thresholdHighStacksFee;

  const mediumFee = shouldCapFee ? appInfo.thresholdHighStacksFee : medium.fee;
  const highFee = shouldCapFee ? appInfo.thresholdHighStacksFee * 1.5 : high.fee;
  const higherFee = fee.multipliedBy(1.25).toNumber();
  const highestFee = fee.multipliedBy(1.5).toNumber();

  const defaultMinimumFee = fee.multipliedBy(1.25).toNumber();
  const minimumFee = !Number.isSafeInteger(defaultMinimumFee)
    ? Math.ceil(defaultMinimumFee)
    : fee.multipliedBy(1.25).toNumber();

  const feePresets: RbfRecommendedFees = fee.lt(BigNumber(mediumFee))
    ? constructRecommendedFees('medium', mediumFee, 'high', highFee, stxAvailableBalance)
    : constructRecommendedFees('higher', higherFee, 'highest', highestFee, stxAvailableBalance);
```

feel free to apply it, if it makes sense

hooks/useRbfTransactionData/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  if (appInfo?.thresholdHighStacksFee) {
+    if (high.fee > appInfo.thresholdHighStacksFee) {
+      // adding a fee cap
+      highFee = appInfo.thresholdHighStacksFee * 1.5;
+      mediumFee = appInfo.thresholdHighStacksFee;
+    }
+  }
+
+  let minimumFee = fee.multipliedBy(1.25).toNumber();
```
fedeerbes, line 92: question: is it okay for the default minimumFee to have the same multiplier as the higherFee here?

hooks/useRbfTransactionData/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  if (fee.lt(BigNumber(mediumFee))) {
+    feePresets = constructRecommendedFees('medium', mediumFee, 'high', highFee, stxAvailableBalance);
+  } else {
+    feePresets = constructRecommendedFees('higher', higherFee, 'highest', highestFee, stxAvailableBalance);
+  }
+
+  return {
+    rbfTransaction: undefined,
```
fedeerbes, line 105: question: do we need to return this here?

hooks/useRbfTransactionData/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+      minimumRbfFee: microstacksToStx(BigNumber(minimumFee)).toNumber(),
+      minimumRbfFeeRate: microstacksToStx(BigNumber(minimumFee)).toNumber(),
+    },
+    rbfRecommendedFees: sortFees(feePresets),
+    mempoolFees: {
+      fastestFee: microstacksToStx(BigNumber(high.fee)).toNumber(),
+      halfHourFee: microstacksToStx(BigNumber(medium.fee)).toNumber(),
+      hourFee: microstacksToStx(BigNumber(slow.fee)).toNumber(),
+      economyFee: microstacksToStx(BigNumber(slow.fee)).toNumber(),
+      minimumFee: microstacksToStx(BigNumber(slow.fee)).toNumber(),
```
fedeerbes, line 118: should we return micro stacks or stx values here? dunno if this util can be used in something else and then you need to do the map to micro stacks again.

hooks/useRbfTransactionData/index.ts
fedeerbes: i think what is confusing here is the name btcNetwork, cause it's actually a settings network, which includes info for stacks and btc. we could rename or make this hook to receive network.type and network.address (stacks url)

fedeerbes: i also think btc rbf logic belong to a separate hook. but i also think that this hook can be two utils, it doesn't need to have logic for these reasons:

* whenever we use this hook fetchRbfData is executed if dependencies change, we kinda loose some flexibility.
* we could handle the logic of data, loading and error in react query in the clients

ignore the idea to be not use hooks, if the reasons are not valid for the current uses cases

hooks/useRbfTransactionData/helpers.ts
teebszet: yup, but do you cover the threshold, or `fee.lt(BigNumber(mediumFee))` code paths? the unit tests should at the least cover the logic which was defined in the notion specs, and ideally more edge cases.

hooks/useRbfTransactionData/helpers.ts
dhriaznov: I think so, according to https://secretkeylabs.slack.com/archives/C063ZVDLCLB/p1703778799805989?thread_ts=1703712686.356469&cid=C063ZVDLCLB and https://secretkeylabs.slack.com/archives/C05QCJM9D1N/p1704972387104769?thread_ts=1704964069.991789&cid=C05QCJM9D1N

hooks/useRbfTransactionData/helpers.ts
dhriaznov: makes sense, thank you, I applied this change

dhriaznov: yeah, the `fee.lt(BigNumber(mediumFee))` code paths were covered there, but I agree that it looked not that obvious, and not easy to read and update in the future.

my initial idea was to test the whole hook, but then I realized that it would be better to make the testing more concise and I like your idea about testing the pure function that is used only for calculation, so I just pushed an update to implement it.

hooks/useRbfTransactionData/helpers.ts
dhriaznov: not sure if this util is going to be used in other circumstances, currently it feels like it's easier to convert it here instead of converting it in both mobile app and web extension in each place where it's used

dhriaznov:
I updated the PR including your suggestions folks, could you take a second look pls? @teebszet @fedeerbes

hooks/transactions/useStxRbfTransactionData.ts
```diff
@@ -0,0 +1,30 @@
...
+    const stxRbfData = await fetchStxRbfData(transaction, btcNetwork, stacksNetwork, appInfo, stxAvailableBalance);
+    return stxRbfData;
+  };
+
+  return useQuery({
+    queryKey: ['stx-rbf-transaction-data', transaction?.txid],
+    queryFn: fetchRbfData,
+    enabled: !!transaction,
+  });
+};
```
teebszet, line 28: so clean!! thank you 🙏 🙏

hooks/useRbfTransactionData/helpers.ts
teebszet: love it, thank you

teebszet:
lgtm 😄

teebszet (approved):

hooks/transactions/useBtcRbfTransactionData.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    return {
+      rbfTransaction: rbfTx,
+      rbfTxSummary: rbfTransactionSummary,
+      rbfRecommendedFees: sortFees(rbfRecommendedFees),
+      mempoolFees,
+    };
+  };
+
+  return useQuery({
```
fedeerbes, line 41: i don't think we should add RQ in core, basically cause add another layer of cache that we need to worry about. but if we do, can we have query config object as an optional param? that way we can customize it in case of need. wdyt? @dhriaznov @teebszet

dhriaznov: @fedeerbes we discussed this question on the daily today, the decision for now is to try to add it since we'll need the same logic on both mobile and web extension, but we're going to be cautious with it and figuring out whether it adds any issues.

PR #377, opened 2024-02-27
Remove nonwitnessutxo injection on pre-sign input preparation

victorkirov:
# 🔘 PR Type

- [x] Bugfix

# 📜 Background
Ledger requires nonWitnessUtxo to be set for segwit inputs in order to not display the non-verified input warning message. However, adding the value results in a PSBT with a slightly different structure, which some integrating apps don't like and they reject the signed PSBT.

# 🔄 Changes

Changes:

- We now only inject the nonWitnessUtxo value when manually adding inputs into transactions. This means that we won't add it when signing external PSBTs.

Impact:

- This should allow you to purchase from marketplaces again

# ✅ Review checklist

### DIFF
--- a/transactions/bitcoin/context.ts
+++ b/transactions/bitcoin/context.ts
@@ -198,37 +198,6 @@ export abstract class AddressContext {
     // this can be implemented by subclasses if they need to do something before signing
   }
 
-  protected async addNonWitnessUtxosToInputs(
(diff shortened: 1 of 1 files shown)


m-aboelenein (approved):

m-aboelenein:
This makes sense. LGTM 👍

